PR #5, opened 2019-11-28
Adjust base path to conventions.

pilhuhn:

### DIFF
--- a/src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java
+++ b/src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java
@@ -45,6 +45,7 @@ import org.testcontainers.containers.PostgreSQLContainer;
 @QuarkusTest
 class RestApiTest {
 
+  private static final String API_BASE = "/api/custom-policies/v1.0";
   @ClassRule
   private static PostgreSQLContainer postgreSQLContainer =
       new PostgreSQLContainer("postgres");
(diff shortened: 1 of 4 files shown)


src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java
```diff
@@ -45,6 +45,7 @@
 @QuarkusTest
 class RestApiTest {
 
+  private static final String API_BASE = "/api/custom-policies/v1.0";
```
josejulio: This is something I have been wondering. 
Do we need to "mount" our backend into `/` and serve the path `/api/custom-policies/v1.0` or do we mount it directly into `/api/custom-policies/v1.0` and we serve all directly from `/` ?

pilhuhn: Well, we should probably not serve `/api/custom-policies/v1.0` from `/` : when there is a need for a v1.1, it may be needed to serve both from `/` for some time (in theory we control the frontend, but first providing v1.1 in parallel allows to independently evolve the client).

Also serving `/api` from `/` would allow to also serve `app/js` from `/` like the Kiali server also serves the UI bits. Not sure if that is needed or not

josejulio: > Well, we should probably not serve `/api/custom-policies/v1.0` from `/` : when there is a need for a v1.1, it may be needed to serve both from `/` for some time (in theory we control the frontend, but first providing v1.1 in parallel allows to independently evolve the client).

My thinking is that both `v1.0` and `v1.1` would be independent binaries thus they won't be in the same "app".

> Also serving /api from / would allow to also serve app/js from / like the Kiali server also serves the UI bits. Not sure if that is needed or not

 I'm not sure, but i think is not needed, as akamai is going to serve all the static bits (?).

PR #6, opened 2019-11-29
Improve policy validation, add policy deletion.

pilhuhn:
This adds better policy validation and persistence handling, along with policy deletion.

Test did fail for me with TestInstantiation errors. See https://github.com/quarkusio/quarkus/issues/5878#issuecomment-559834520. This is fixed with Quarkus 1.0


pilhuhn:
Removed the DNM label as with Quarkus 1.1 the test instantiation issues are gone.

pilhuhn:
Updated to follow changes from master.

src/main/java/com/redhat/cloud/custompolicies/app/model/FactType.java
```diff
@@ -23,6 +23,7 @@
 
   STRING("STRING"),
   BOOLEAN("BOOLEAN"),
+  LIST("LIST"),
```
burmanm: Just a nitpick, but aren't these arrays actually (in JSON) ?

pilhuhn:
@theute Do you recall why you set the DNM label?

theute:
> @theute Do you recall why you set the DNM label?

I think it was waiting on another change, you can definitely remove if it's ready to merge

src/main/java/com/redhat/cloud/custompolicies/app/model/FactType.java
pilhuhn: For the moment this is meant as a hint to the UI what datatype a fact is, so that it can show the right set of comparators.
I can change it to array, but I don't think it really matters.

pilhuhn:
This is ready from my side. I updated the original comment to mention that the test failure is gone with Quarkus 1.1

theute:
Having a look

theute:
I'm unable to build the engine at this time (and that led me to open this issue: https://issues.redhat.com/browse/CPOL-41 ) but otherwise seems good.
@burmanm I let you merge if you have no further comment

PR #7, opened 2019-12-03
Allow building container-images.

pilhuhn:
Allow to build container images.
This is off of master and may/will clash with #6 so one or the other will need rebasing

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -1,0 +1,19 @@
+FROM registry.access.redhat.com/ubi7-minimal
+
+LABEL maintainer="hrupp@redhat.com"
+RUN microdnf install java-1.8.0-openjdk-headless && microdnf clean all
+
+ENV quarkus.http.port=8080 \
+    quarkus.datasource.url=jdbc:postgresql://localhost:5432/postgres \
+    quarkus.datasource.username=sarah \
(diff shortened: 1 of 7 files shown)


deploy/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM registry.access.redhat.com/ubi7-minimal
```
theute: Any reason why we don't use UBI8 ?

deploy/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+LABEL maintainer="hrupp@redhat.com"
+RUN microdnf install java-1.8.0-openjdk-headless && microdnf clean all
```
theute: And java-11-openjdk-headless ?

deploy/Dockerfile
pilhuhn: I implicitly associated os311 with rhel7 and ubi7. We probably should investigate this.

deploy/Dockerfile
pilhuhn: No specific reason.

deploy/Dockerfile
theute: So since we're running on OS3/RHEL7, the best supported container host would be UBI7.
My question got answered

PR #10, opened 2019-12-11
Use a JaxRs application class to provide the base path.

pilhuhn:


pilhuhn:
Will send a new one without the logging change and then logging in yet another one.

Issue #13, opened 2019-12-17
OpenApi path should not be inside the normal rest resources


pilhuhn:
https://github.com/RedHatInsights/custom-policies-ui-backend/blob/39ae6eae63d7f5be9dd6456f3f4c5160e3d91721/src/main/resources/application.properties#L10

Remove the last 'policies'

PR #14, opened 2019-12-17
Fixed openapi.json URL according to the standardized location

theute:
That seems to fix the openapi.json file to be served from the right URL


pilhuhn:
This is about #13

PR #15, opened 2019-12-17
Ensure that the openapi document is served at the right place.

pilhuhn:
Followup to #13, #14


theute:
I should have done that, thanks !

PR #16, opened 2019-12-17
Produce a structured access log (goes to stdout).

pilhuhn:
Output looks like 

`{"date":"Tue, 17 Dec 2019 15:31:03 GMT","method":"GET","uri":"/api/custom-policies/v1.0/facts","http_version":"HTTP/1.1","status":401,"content_length":0,"duration":83,"user_agent":"curl/7.54.0","remote":"0:0:0:0:0:0:0:1"}
`

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/JaxRsApplication.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/JaxRsApplication.java
@@ -16,6 +16,10 @@
  */
 package com.redhat.cloud.custompolicies.app;
 
+import io.vertx.core.Handler;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+import javax.enterprise.event.Observes;
 import javax.ws.rs.ApplicationPath;
 import javax.ws.rs.core.Application;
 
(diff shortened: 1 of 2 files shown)


burmanm:
Shouldn't the log pattern make it JSON instead? The logging itself should only make properties and the logging library should modify it to the wanted output.

pilhuhn:
Well, that exists for "normal logging". Piping the access log into this ends up with something like

```
{"timestamp":"2019-12-13T16:52:56.929+01:00",
 "sequence":2276,
 "loggerClassName":"java.util.logging.Logger",
 "loggerName":"com.redhat.cloud.custompolicies.app.JsonLoggerHandler",
 "level":"INFO",
 "message":"{\"date\":\"Fri, 13 Dec 2019 15:52:56 GMT\",\"duration\":18,\"method\":\"GET\",\"content_lenght\":2,\"http_version\":\"HTTP/1.1\",\"remote\":\"0:0:0:0:0:0:0:1\",\"uri\":\"/api/custom-policies/v1.0/test\",\"user_agent\":\"curl/7.54.0\",\"status\":200}",
 "threadName":"vert.x-worker-thread-6",
 "threadId":132,
 "mdc":{},
 "ndc":"",
 "hostName":"snert",
 "processName":"app-dev.jar",
 "processId":67668
}

```

where the message field contains the json blob in some heavy quoted way.

PR #17, opened 2019-12-17
Use/allow jdk11

pilhuhn:


pilhuhn:
Once container stuff is merged, this needs to update the base image to 11 as well.

pilhuhn:
We discussed this and said to proceed, so removing the DNM

pilhuhn:
Update of container image is no longer needed, as #7 was merged.

PR #18, opened 2019-12-19
CPOL-32 Remove customerid from urls

pilhuhn:
Customeids in urls was an early stop-gap that allowed to select policy sets for different users.
Now that the id is supplied via x-rh-identity header, we can remove it.

https://issues.redhat.com/browse/CPOL-32

@zsadeh

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/model/Policy.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/model/Policy.java
@@ -17,9 +17,9 @@
 package com.redhat.cloud.custompolicies.app.model;
 
 import io.quarkus.hibernate.orm.panache.PanacheEntity;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.validation.constraints.NotEmpty;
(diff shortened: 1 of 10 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/model/Policy.java
```diff
@@ -44,6 +44,23 @@
   public String conditions;
   public String actions;
 
+  // De-serialise the array form of actions coming from the UI
+  // this needs more work with some more typesafe translations
+  // or alternatively storing of actions as json blob
+  // See CPOL-33
+  public void setActions(List<Map<String, String>> actionList) {
+    System.out.println(actionList);
```
burmanm: Forgotten debug!

src/main/java/com/redhat/cloud/custompolicies/app/model/Policy.java
```diff
@@ -44,6 +44,23 @@
...
+    Iterator<Map<String,String>> ite = actionList.iterator();
+    actions = "";
```
burmanm: Wouldn't StringBuilder make more sense here? Otherwise, the code would be quite a bit more simple I guess with something like:

```java
actionList.stream().map(m -> m.get("type")).collect(Collectors.join(", "));
```

Or 

```
StringJoiner joiner = new StringJoiner(", ");
for(Map<String, String map> map : actionList) {
joiner.add(map.get("type"));
}
```

Just all the tricks in the Java 8+ world ;)

pilhuhn: Yes it would. I see this code as very temporary -- see comment about CPOL-33.
But I can adopt your suggestions anyway

PR #19, opened 2020-01-13
CPOL-42 Allows pagination

josejulio:
~~I'm adding `lombok`, it needs IDE configuration to function properly:~~

~~Eclipse: https://projectlombok.org/setup/eclipse
Intelli: https://projectlombok.org/setup/intellij~~

Removed it

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,6 +119,12 @@
       <version>3.8.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
+      <version>1.18.10</version>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -119,12 +119,6 @@
       <version>3.8.0</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.projectlombok</groupId>
-      <artifactId>lombok</artifactId>
-      <version>1.18.10</version>
-      <scope>provided</scope>
-    </dependency>
   </dependencies>
   <build>
     <plugins>
--- a/src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
@@ -72,7 +72,7 @@ public class PagingUtils {
         ResponseBuilder builder;
 
         if (page.isEmpty()) {
-            builder = Response.status(Response.Status.NO_CONTENT);
+            builder = Response.status(Response.Status.NOT_FOUND);
         } else {
             builder = Response.ok(page);
             EntityTag etag = new EntityTag(String.valueOf(page.hashCode()));
--- a/src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
@@ -36,16 +36,33 @@ public class PagingUtils {
         Pager.PagerBuilder pageBuilder = Pager.builder();
         MultivaluedMap<String, String> queryParams = uriInfo.getQueryParameters();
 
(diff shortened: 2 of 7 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static Pager extractPager(@NotNull UriInfo uriInfo) {
+        Pager.PagerBuilder pageBuilder = Pager.builder();
+        MultivaluedMap<String, String> queryParams = uriInfo.getQueryParameters();
+
+        try {
+            pageBuilder.page(Integer.parseInt(queryParams.getFirst("page")));
+        } catch (NumberFormatException nfe) {
+            // Ignore the exception, the builder will use the default value
```
pilhuhn: Throwing and catching exceptions  is expensive. Better check if a parameter is present

src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        try {
+            pageBuilder.itemsPerPage(Integer.parseInt(queryParams.getFirst("pageSize")));
+        } catch (NumberFormatException nfe) {
+            // Ignore the exception, the builder will use the default value
```
pilhuhn: Same as above

pom.xml
```diff
@@ -119,6 +119,12 @@
       <version>3.8.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.projectlombok</groupId>
+      <artifactId>lombok</artifactId>
```
pilhuhn: I don't think this is a good idea. This pulls in a large lib for little gain.

src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
burmanm: In the context of this application, Exception is cheap (we're talking about hundreds of millions of ops per second). However, there's no replacement for Exception here - the parseInt can throw Exception if the user sends "page=second".

The question is rather - should we really use the default or return this as error to the user?

josejulio: > However, there's no replacement for Exception here - the parseInt can throw Exception if the user sends "page=second".

That's the use there, right.

> The question is rather - should we really use the default or return this as error to the user?

Good point.
I'll better throw the exception to let the user know about this error.

pom.xml
josejulio: I was more interested on the `Builder`, looks more DRY and is configurable as simple annotations. 
But I agree that is a little gain for now. I'll remove the lib.

josejulio:
Updated the code: 
 - no more `lombok`, added the builder and getters.
 - NFE are reported to the client, but still using defaults if page or pagePerItem are not passed.

src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
```diff
@@ -0,0 +1,85 @@
...
+    public static <T>ResponseBuilder responseBuilder(Page<T> page) {
+        ResponseBuilder builder;
+
+        if (page.isEmpty()) {
+            builder = Response.status(Response.Status.NOT_FOUND);
+        } else {
+            builder = Response.ok(page);
+            EntityTag etag = new EntityTag(String.valueOf(page.hashCode()));
+            builder.header("ETag",etag);
+            builder.header("TotalCount", Long.toString(page.getTotalCount()));
```
pilhuhn: I don't know for sure, but  I think custom headers should have a `X-` prefix. Doesn't really harm leaving it as it I guess

josejulio: I was reading that the `X-` prefix has been deprecated: https://tools.ietf.org/html/rfc6648

PR #21, opened 2020-01-15
Do DB-schema setup/migrations at application start.

pilhuhn:

cc @zsadeh


src/main/resources/db/migration/V1.sql
```diff
@@ -0,0 +1,21 @@
...
+create table policy (id integer,
+                     customerid varchar not null,
+                     name varchar not null ,
+                     description varchar,
+                     is_enabled boolean not null default false,
+                     conditions varchar not null,
+                     actions varchar );
+
+alter table policy ADD PRIMARY KEY (customerid,id);
+create unique index pol_cus_nam_idx on policy (customerid,name); -- TODO does this make the name unique per customer (?)
```
josejulio: It does:
```
=# insert into policy values (99, '940527', '1st policy', 'Just a test', true, '"cores" == 1','EMAIL roadrunner@acme.org');
ERROR:  duplicate key value violates unique constraint "pol_cus_nam_idx"
DETAIL:  Key (customerid, name)=(940527, 1st policy) already exists.

```

PR #23, opened 2020-01-16
Improve documentation.

pilhuhn:

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -49,12 +49,14 @@ curl  -Hx-rh-identity:`cat src/test/resources/rhid.txt` http://localhost:8080/ap
 By default this uses paging with a page-size of 10.
 You can select different pages and page sizes with the following query parameters
 
+|===
+|Parameter|Default|Comment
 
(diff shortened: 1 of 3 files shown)


README.adoc
```diff
@@ -31,26 +31,69 @@ When developing you can use Quarkus hot-reload
...
 
+By default this uses paging with a page-size of 10.
+You can select different pages, page sizes and sorting with the following query parameters:
+
+|===
+|Parameter|Default|Comment
+
+|page     | 0     | Page numbers start at 0
+|pageSize | 10    |
+|sortColumn |  | Sort can happen by column names ("name",
```
pilhuhn: @josejulio Is(n't) default sorting by internal id ?

josejulio: mmmh, Not sure what postgres does by default, but it looks like that, yes.

pilhuhn: It looks like default is by id (primary key) column. Should we perhaps make the name the default sort argument (unless we get the 'last updated' column)?

PR #24, opened 2020-01-16
CPOL-43 Allows sorting for policies

josejulio:


src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
```diff
@@ -65,6 +69,28 @@ public static Pager extractPager(@NotNull UriInfo uriInfo) {
             }
         }
 
+        List<String> columns = queryParams.get(QUERY_COLUMN);
+        List<String> directions = queryParams.get(QUERY_DIRECTION);
+        if (columns != null) {
+            for (int i = 0; i < columns.size(); ++i) {
+                String column = columns.get(i);
```
pilhuhn: I think those need to be sanitized. Not sure. Perhaps we should either reject unknown column names. Or pass something like an index and then internally create the column name via that index (actually in both cases an Enum with column names could come handy)

src/main/java/com/redhat/cloud/custompolicies/app/model/Policy.java
```diff
@@ -47,7 +48,12 @@
 
 
   public static Page<Policy> pagePoliciesForCustomer(String customer, Pager pager) {
-    PanacheQuery<Policy> query = find("customerid", customer)
+
+    for (Sort.Column column : pager.getSort().getColumns()) {
+      SortableColumn.fromName(column.getName());
+    }
```
josejulio: @pilhuhn Is this what you had in mind? 
Other alternative that i was thinking is to add an object describing what the `PagingUtils` should allow, e.g.
Allow sorting by a, b and c, allow filtering of b, etc.
And so the RestService would be in charge of describing their needs depending on the service and maybe the user access.

pilhuhn: I think this is good now.

josejulio: OK, I'll fix the CI error to make travis happy again.

pilhuhn:
Can you please add test in RestApiTest, that 
- try sorting
- check that a bad column name passed results in an error to the client?

josejulio:
> Can you please add test in RestApiTest, that
> 
> * try sorting
> * check that a bad column name passed results in an error to the client?

Yes, will do.

josejulio:
> Can you please add test in RestApiTest, that
> 
> * try sorting
> * check that a bad column name passed results in an error to the client?

Done

PR #25, opened 2020-01-17
Fix generation of Location header and test this.

pilhuhn:
This was missing a path segment, as the api of UriBuilder is a bit cryptic


src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java
```diff
@@ -195,7 +195,23 @@ void storeNewPolicy() {
     assert headers.hasHeaderWithName("Location");
     // TODO extract id and then check in subsequent call
     // that the policy is stored
```
josejulio: I think you already did this on this PR :)

pilhuhn: Was only checking if someone pays attention :-)

PR #28, opened 2020-01-20
Actually send the policy to the engine when it is to be stored.

pilhuhn:
Also introduces a flag to skip the engine calls.
  Mostly meant for local development.

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
@@ -45,6 +45,7 @@ import javax.ws.rs.core.UriInfo;
 import com.redhat.cloud.custompolicies.app.model.pager.Page;
 import com.redhat.cloud.custompolicies.app.model.pager.Pager;
 import com.redhat.cloud.custompolicies.app.rest.utils.PagingUtils;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.metrics.annotation.Timed;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.enums.ParameterIn;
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
```diff
@@ -126,13 +130,14 @@ public Response storePolicy(@QueryParam ("alsoStore") boolean alsoStore, @Valid
...
-    catch (Exception e) {
-      msg = new Msg(e.getMessage());
-      return Response.status(400,e.getMessage()).entity(msg).build();
+    if (!skipEngineCall) {
+      try {
+        FullTrigger trigger = new FullTrigger(policy);
+        engine.store(trigger, true, user.getAccount());
+      } catch (Exception e) {
+        msg = new Msg(e.getMessage());
+        return Response.status(400, e.getMessage()).entity(msg).build();
```
josejulio: Trying to store a second policy yields:

```{"msg":"Validation failed: Trigger with ID [defaultTenant] exists."}```

Probably a engine issue?

josejulio: Mmh, not sure why started happening when testing this PR, but I've tried to clean my environment (docker-compose down, docker volume prune) and run off master and still the same message.

Running with ENGINE_SKIP=true allowed me to continue working.

@burmanm do you know what could be wrong? Am I missing a setup step?

burmanm: No, that is correct error message - the trigger already exists with the id.

burmanm: The current dry (validation) method checks that, since it's a validation error if one later tries to store it.

josejulio: Mmh, the backend is the one that should assign the id, but at this point we only want to know if the condition is OK

Should we Skip the id validation when is a dry run (when is the default id or if the id is not present) ?

josejulio: OK, after reading the code I noticed that the ID is by default an UUID.
I'll create one in the backend code and pass to the engine.

josejulio:
Rebased to master

PR #31, opened 2020-01-22
Add mtime column and make sort default to 'mtime'

pilhuhn:
Explicitly sort by name if no sort column is given.
Add 'last updated' column with the name 'mtime'.


josejulio:
There is an error with travis, and not sure what to make of it, previously it worked different.

In [RestApiTest](https://github.com/RedHatInsights/custom-policies-ui-backend/blob/1e9e0f79ae0f0e9670dfd054660508e13a2363fe/src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java#L158). I query a service that throws an exception, but I expect that the client gets an error 500 (as in the test), but now is throwing the exception from the server.

josejulio:
> Explicitly sort by name if no sort column is given.

I'm not sure if name would be the best default, but I don't really mind. I think the UI would default to mtime DESC, to show latests policies at first.

PR #37, opened 2020-02-14
CPOL-83 Return stored policy in response.

pilhuhn:

cc @zsadeh

### DIFF
--- a/src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java
+++ b/src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java
@@ -28,6 +28,7 @@ import io.restassured.http.ContentType;
 import io.restassured.http.Header;
 import io.restassured.http.Headers;
 import io.restassured.path.json.JsonPath;
+import io.restassured.response.ExtractableResponse;
 import javax.json.bind.Jsonb;
 import javax.json.bind.JsonbBuilder;
 import org.junit.Assert;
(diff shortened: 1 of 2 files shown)


src/test/java/com/redhat/cloud/custompolicies/app/RestApiTest.java
```diff
@@ -307,6 +314,7 @@ void storeNewPolicy() {
 
     assert body.get("conditions").equals("cores = 2");
     assert body.get("name").equals("test1");
+    assert (Integer)body.get("id") == returnedBody.id;
```
burmanm: Hopefully this isn't autoincrement integer?

pilhuhn: It may be under the covers, but JPA is taking care of it. 
If this int would only be incremented after the store being completed, this assert would fail.
Also GETting the policy by id (location) would fail before.

burmanm: That was not the point, but the fact that it reveals potentially too much. For example if the service is not used by many customers, it can be easily seen from the new id one generates. I'm not sure if that's wanted behavior.

pilhuhn: We can change that - can you please open a JIRA for it? The change should not be part of this PR (imo), as this is about something else.

pilhuhn: I have opened CPOL-92 for the Id change to UUID.

PR #39, opened 2020-02-14
CPOL-44 Adds filtering to Policy

josejulio:


josejulio:
> Shouldn't there some update to the OpenAPI for getPolicies()?
> Also can you add the filter options to the readme please?

Ah, I think so, the new request params.

PR #40, opened 2020-02-17
CPOL-84 implement paging according to platform standards.

pilhuhn:
This requires changes in the UI as well. 

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
@@ -16,6 +16,8 @@
  */
 package com.redhat.cloud.custompolicies.app.rest.utils;
 
+import static java.lang.Long.max;
+
 import com.redhat.cloud.custompolicies.app.model.Policy;
 import com.redhat.cloud.custompolicies.app.model.pager.Page;
 import com.redhat.cloud.custompolicies.app.model.pager.Pager;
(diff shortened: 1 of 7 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/rest/utils/PagingUtils.java
```diff
@@ -41,8 +45,8 @@ public static Pager extractPager(@NotNull UriInfo uriInfo) {
         Pager.PagerBuilder pageBuilder = Pager.builder();
         MultivaluedMap<String, String> queryParams = uriInfo.getQueryParameters();
 
-        final String QUERY_PAGE = "page";
-        final String QUERY_PAGE_SIZE = "pageSize";
+        final String QUERY_PAGE = "offset";
```
josejulio: Actually, i don't think the `pageSize` and `offset` are the same, per the [pagination docs](https://platform-docs.cloud.paas.psi.redhat.com/rest/pagination.html#requests) 
> offset indicates the starting position of the query relative to the complete set of items that match the client’s query.

This is actually the number of elements that we are going to "ignore", not the page we want.
You can check their examples to confirm this, see how the first example has `last` and `next` as `offset=10` when the `count=11`.
https://platform-docs.cloud.paas.psi.redhat.com/rest/pagination.html#examples

pilhuhn: You are right. Good catch

PR #41, opened 2020-02-18
CPOL-89 Allows to dry run when updating a policy

josejulio:


src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
```diff
@@ -238,29 +238,22 @@ public Response storePolicy(@QueryParam ("alsoStore") boolean alsoStore, @Valid
     policy.id = null;
     policy.customerid = user.getAccount();
 
-    Msg msg ;
+    Policy tmp = Policy.findByName(user.getAccount(), policy.name);
+    if (tmp != null) {
+      return Response.status(409).entity(new Msg("Policy name is not unique")).build();
```
josejulio: Moved this validation over here to allow this to happen even when we intend to save the policy.

PR #45, opened 2020-02-24
CPOL-63 send actions to engine

pilhuhn:

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
@@ -17,7 +17,11 @@
 package com.redhat.cloud.custompolicies.app.model;
 
 import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.UUID;
 
 /**
@@ -51,6 +55,26 @@ public class FullTrigger {
     var cond = new Condition();
     cond.expression = policy.conditions;
     conditions.add(cond);
+    storeActions(trigger, policy);
+  }
+
+  private void storeActions(Trigger trigger, Policy policy) {
+    String[] actionsIn = policy.actions.split(";");
+    for (String actionIn : actionsIn) {
+      TriggerAction ta = new TriggerAction();
+      if (actionIn.startsWith("EMAIL")) {
+        ta.actionPlugin = "email";
+      } else if (actionIn.startsWith("HOOK")) { // TODO what does the UI send?
+        ta.actionPlugin = "webhook";
+        String endpoint = actionIn.substring(actionIn.indexOf(' ')+1);
+        ta.properties.put("endpoint_id",endpoint);
+      }
+      else {
+        throw new IllegalArgumentException("Unknown action type " + actionIn);
+      }
+
+      trigger.actions.add(ta);
+    }
   }
 
   public FullTrigger(Policy policy) {
@@ -67,7 +91,7 @@ public class FullTrigger {
     public String name;
     public String description;
     public boolean enabled;
-
+    public Set<TriggerAction> actions;
 
   }
 
@@ -76,5 +100,10 @@ public class FullTrigger {
     public String type = "EVENT";
     public String dataId = "insights_report";
     public String expression;
+  }
+
+  public static class TriggerAction {
+    public String actionPlugin;
+    public Map<String,Object> properties = new HashMap<>();
   }
 }
--- a/src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
@@ -59,12 +59,19 @@ public class FullTrigger {
   }
 
   private void storeActions(Trigger trigger, Policy policy) {
+    if (policy.actions == null) {
+      return;
+    }
+
     String[] actionsIn = policy.actions.split(";");
     for (String actionIn : actionsIn) {
       TriggerAction ta = new TriggerAction();
-      if (actionIn.startsWith("EMAIL")) {
+      if (actionIn.trim().isEmpty()) {
+        continue;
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
```diff
@@ -51,6 +56,26 @@ public FullTrigger(Policy policy) {
     Condition cond = new Condition();
     cond.expression = policy.conditions;
     conditions.add(cond);
+    storeActions(trigger, policy);
+  }
+
+  private void storeActions(Trigger trigger, Policy policy) {
+    String[] actionsIn = policy.actions.split(";");
```
josejulio: This will fail if `actions` is `null`. If is empty string, it will throw later as this will yield an array of one element `""`

src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
```diff
@@ -51,6 +56,26 @@ public FullTrigger(Policy policy) {
...
+    for (String actionIn : actionsIn) {
+      TriggerAction ta = new TriggerAction();
+      if (actionIn.startsWith("EMAIL")) {
```
josejulio: We actually use lower case for the action type:
https://github.com/RedHatInsights/custom-policies-ui-frontend/blob/cb562622c4cceb47b0ff2b898f112f6ac801e80c/src/types/Policy/Actions/ActionType.ts#L2-L3

src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
pilhuhn: Good catch!

src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
```diff
@@ -51,6 +56,26 @@ public FullTrigger(Policy policy) {
...
+        ta.actionPlugin = "email";
+      } else if (actionIn.startsWith("HOOK")) { // TODO what does the UI send?
```
josejulio: UI sends WEBHOOK

pilhuhn: So upper case here?  :-)  *joking*

josejulio: whoops :), `webhook`

src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
```diff
@@ -51,6 +56,33 @@ public FullTrigger(Policy policy) {
...
+    String[] actionsIn = policy.actions.split(";");
+    for (String actionIn : actionsIn) {
+      TriggerAction ta = new TriggerAction();
+      if (actionIn.trim().isEmpty()) {
+        continue;
+      }
+      if (actionIn.toLowerCase().startsWith("email")) {
+        ta.actionPlugin = "email";
+      } else if (actionIn.toLowerCase().startsWith("webhook")) {
+        ta.actionPlugin = "webhook";
```
pilhuhn: Needs to be changed to 'hooks' for the old hooks. *web*hooks is too modern :)

src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
```diff
@@ -65,14 +65,14 @@ private void storeActions(Trigger trigger, Policy policy) {
...
-      TriggerAction ta = new TriggerAction();
+      var ta = new TriggerAction();
       if (actionIn.trim().isEmpty()) {
         continue;
       }
       if (actionIn.toLowerCase().startsWith("email")) {
         ta.actionPlugin = "email";
       } else if (actionIn.toLowerCase().startsWith("webhook")) {
-        ta.actionPlugin = "webhook";
+        ta.actionPlugin = "hook"; // The legacy hooks apps
```
burmanm: It's hooks with ``s`` at the end ;)

https://github.com/RedHatInsights/custom-policies-engine/blob/master/external/src/main/java/com/redhat/cloud/custompolicies/engine/actions/plugins/BetaHooksActionPluginRegister.java#L26

src/main/java/com/redhat/cloud/custompolicies/app/model/FullTrigger.java
```diff
@@ -47,10 +52,36 @@ public FullTrigger(Policy policy, boolean generatePseudoId) {
...
+      var ta = new TriggerAction();
+      if (actionIn.trim().isEmpty()) {
+        continue;
+      }
+      if (actionIn.toLowerCase().startsWith("email")) {
+        ta.actionPlugin = "email";
+      } else if (actionIn.toLowerCase().startsWith("webhook")) {
+        ta.actionPlugin = "hooks"; // The legacy hooks apps
+        String endpoint = actionIn.substring(actionIn.indexOf(' ')+1);
+        ta.properties.put("endpoint_id",endpoint);
```
burmanm: Right, I did not notice this on our review session. But the old webhooks application does not work this way, we can't define the target destination. Instead, the target destination is all the activated webhooks in the old webhooks app.  So there's no parameters at all.

josejulio: Not even an id or something to identify which webhook we want to call?

theute: Oh that's a good point, it doesn't look like we can specify which webhook to trigger in the end  :man_facepalming: https://github.com/RedHatInsights/notifications-backend
CC: @jeromemarc

burmanm: @josejulio Nope, nothing. That's one of my gripes with it, it adds another "rules engine" on top of the existing and has just bad rules.

It's simply all or nothing for the old one.

PR #48, opened 2020-02-27
CPOL-104 Implement endpoint for user email preferences

pilhuhn:

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/NotificationSystem.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/NotificationSystem.java
@@ -33,16 +33,14 @@ public interface NotificationSystem {
 
 
   @PUT
-  @Path("/{userId}/{event}")
+  @Path("/{event}")
   CompletionStage<String> addNotification(
-      @PathParam("userId") String userId,
       @PathParam("event") String event,
       @HeaderParam("x-rh-identity") String rhIdentity);
 
   @DELETE
-  @Path("/{userId}/{event}")
+  @Path("/{event}")
   CompletionStage<String> removeNotification(
-      @PathParam("userId") String userId,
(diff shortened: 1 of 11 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/NotificationSystem.java
```diff
@@ -0,0 +1,48 @@
...
+@RegisterRestClient(configKey = "notifications")
+public interface NotificationSystem {
+
+
+  @PUT
+  @Path("/{userId}/{event}")
+  CompletionStage<String> addNotification(
+      @PathParam("userId") String userId,
+      @PathParam("event") String event,
+      @HeaderParam("x-rh-identity") String rhIdentity);
```
burmanm: If you push the x-rh-identity, we can leave out the userId part. So ``/{event}`` would be enough.

src/main/java/com/redhat/cloud/custompolicies/app/rest/SettingsService.java
```diff
@@ -0,0 +1,125 @@
...
+  RhIdPrincipal user;
+
+  @Inject
+  @RestClient
+  NotificationSystem notifications;
+
+  @Operation(summary = "Save or update settings from the settings UI")
+  @APIResponse(responseCode = "200", description = "Saving was ok")
+  @POST
+  @Path("/email-preference")
```
josejulio: We might need to change this so that the endpoint is `/settings` in both load/save.
Asking Karel, but already got WIP to have this working. Will send a PR to your branch later today.

pilhuhn: I talked to Karel and he told me I did the right thing

PR #50, opened 2020-02-28
CPOL-92 Use a UUID as PK for Policies.

pilhuhn:
Also use that UUID as trigger id.

WARNING: needs a re-set of the database. Migration is not incremental

This also needs some changes on the UI side.


pilhuhn:
Reviewed offline at the F2F

PR #52, opened 2020-03-06
CPOL-92 Don't try to install the UUID extension.

pilhuhn:
Our user does not have enough rights in the DB.
And it was only used in tests, so remove and use
explicit uuids.


zsadeh:
Looks good.

PR #61, opened 2020-03-13
Add support for distributed tracing

pilhuhn:
This is a larger change. I tried this with Quarkus 1.2 but failed (most likely pilot error), so I pulled in 1.3 where I got it to work.

Bildschirmfoto 2020-03-13 um 11 26 38

- bump Quarkus to 1.3.0.Final
  +- Change test setup to use QuarkusTestResource
- Pull in MP-OpenTracing and set it up in application.properties.
- Separate out Rbac retrieval from IncomingRequestFilter, as otherwise it would not be part of the outer span (that only starts when the endpoint is hit, but the IRF is a @PreMatch filter, which runs before
- Implement a Rbac cache: This is a simple annotation that creates the cache. Entries expire after 2mins (see application.properties) 

With engine also instrumented :

Bildschirmfoto 2020-03-13 um 12 42 21

See also https://github.com/RedHatInsights/custom-policies-engine/pull/42

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/auth/IncomingRequestFilter.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/auth/IncomingRequestFilter.java
@@ -39,13 +39,8 @@ import javax.ws.rs.ext.Provider;
  * principal will also be made available for Injection, so that you can write in
  * your code.
  *
-<<<<<<< HEAD
- * We don't yet query for RBAC here, but in its own filter after the
- * JAX-RS engine has selected the endpoint to call.
-=======
  * We don't yet query for RBAC here, as this filter is not part of the tracing
  * span, so we would not be able to trace the rbac calls.
->>>>>>> Add support for distributed tracing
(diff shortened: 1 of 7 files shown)


pilhuhn:
Note: I pulled the entire team in as reviewer just as a FYI

src/main/java/com/redhat/cloud/custompolicies/app/auth/IncomingRequestFilter.java
```diff
@@ -39,8 +39,13 @@
  * principal will also be made available for Injection, so that you can write in
  * your code.
  *
+<<<<<<< HEAD
  * We don't yet query for RBAC here, but in its own filter after the
  * JAX-RS engine has selected the endpoint to call.
+=======
+ * We don't yet query for RBAC here, as this filter is not part of the tracing
+ * span, so we would not be able to trace the rbac calls.
+>>>>>>> Add support for distributed tracing
```
josejulio: Some conflict fixing leftovers.

PR #63, opened 2020-03-17
CPOL-66 Rbac caching

pilhuhn:
Depends on #62 (actually it includes it)

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/auth/RbacFilter.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/auth/RbacFilter.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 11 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/auth/RbacFilter.java
```diff
@@ -0,0 +1,74 @@
...
+    try {
+      result = getRbacInfo(user.getRawRhIdHeader());
+    } catch (Throwable e) {
+      requestContext.abortWith(Response.status(Response.Status.FORBIDDEN).build());
+      return;
+    }
+
+    user.setRbac(result.canReadAll(),result.canWriteAll());
+    RhIdPrincipal userPrincipal = (RhIdPrincipal) requestContext.getSecurityContext().getUserPrincipal();
+    userPrincipal.setRbac(result.canReadAll(),result.canWriteAll());
```
josejulio: Why it adds it to `user` and then a `userPrincipal`? What's the difference between those two objects?

pilhuhn: The one is stored in the security context and the other made available via injection.
Currently the code only uses injection, but I make both available for convenience.
Btw. That was already in the previous implementation. It only was now pulled out into a separate class.

PR #65, opened 2020-03-18
Harden settings service.

pilhuhn:
Especially when notification system can't be reached.

cc @karelhala

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/rest/SettingsService.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/rest/SettingsService.java
@@ -57,10 +57,18 @@ public class SettingsService {
 
   @Operation(summary = "Save or update settings from the settings UI")
   @APIResponse(responseCode = "200", description = "Saving was ok")
+  @APIResponse(responseCode = "403", description = "User has no permission to change settings")
+  @APIResponse(responseCode = "500", description = "Saving of settings failed")
   @POST
   @Path("/")
   @Transactional
   public Response saveSettings(@Valid SettingsValues values) {
+
+    Response.ResponseBuilder builder;
+
+    if (!user.canWriteAll()) {
+      return Response.status(Response.Status.FORBIDDEN).entity("You don't have permission to change settings").type("text/plain").build();
+    }
 
     values.username = user.getName();
     values.accountId = user.getAccount();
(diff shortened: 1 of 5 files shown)


theute:
Unable to do a real review, but definitely review to give it a try once merged

src/main/java/com/redhat/cloud/custompolicies/app/rest/SettingsService.java
```diff
@@ -57,43 +57,64 @@
...
+
     values.username = user.getName();
     values.accountId = user.getAccount();
     SettingsValues tmp = SettingsValues.findById(user.getName());
     if (tmp != null) {
       tmp.immediateEmail = values.immediateEmail;
       tmp.dailyEmail = values.dailyEmail;
     } else {
-      values.persistAndFlush();
+      values.persist();
```
burmanm: This still persists changes even if the store fails?

pilhuhn: Should not, as there is a `@Transactional` annotation

burmanm: Line 97 has ``catch (Exception e) {`` thus the JTA doesn't get any exception. So, what would rollback?

src/main/java/com/redhat/cloud/custompolicies/app/rest/SettingsService.java
```diff
@@ -57,43 +57,64 @@
...
+        notifications.removeNotification("custom-policies-daily-mail", user.getRawRhIdHeader());
+      }
+      builder = Response.ok();
     }
-    else {
-      notifications.removeNotification("custom-policies-daily-mail", user.getRawRhIdHeader());
+    catch (Exception e) {
+      builder = Response.serverError().entity("Storing of settings in notification service failed.");
+      builder.type("text/plain");
+      System.err.println("Storing settings failed: " + e.getMessage());
```
burmanm: Should probably be a logger line?

src/main/resources/application.properties
```diff
@@ -27,7 +27,7 @@ engine.skip=false
 rbac/mp-rest/url=https://ci.cloud.redhat.com
 
 # Notifications server
-notifications/mp-rest/url=http://custom-policies-notifications:8080
+notifications/mp-rest/url=http://custom-policies-notifications:8000
```
theute: How about we expose port 8080 like for other containers ?
I guess we just need to change:
https://github.com/RedHatInsights/e2e-deploy/blob/master/templates/custom-policies/custom-policies-notifications.yml#L15

theute: I made a PR here: https://github.com/RedHatInsights/e2e-deploy/pull/1242
It that gets approved, I can already test I think

pilhuhn: I am not opposed to that. Having 8080 everywhere makes it harder locally when all services are deployed 'bare metal'. cc @burmanm

karelhala:
Error codes and return values look good to me!

src/main/resources/application.properties
burmanm: Well, container can redirect ports so it doesn't matter - that PR didn't change the local port.

theute: So it's running on port 8080 now, we can change back if you prefer.

pilhuhn: What is the conclusion here? Do I change the default in here or not?

theute: it's listening on 8080 now

src/main/java/com/redhat/cloud/custompolicies/app/rest/SettingsService.java
pilhuhn: You are right. I changed this now.

PR #69, opened 2020-03-19
CPOL-63 (followup)

pilhuhn:
Fix the update logic.

There is one part of a test currently disabled, as this needs more work in the mockserver to correctly identify the policy and then reply with a correctly mocked FullTrigger for further processing.

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/model/engine/FullTrigger.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/model/engine/FullTrigger.java
@@ -14,8 +14,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.redhat.cloud.custompolicies.app.model;
+package com.redhat.cloud.custompolicies.app.model.engine;
 
+import com.redhat.cloud.custompolicies.app.model.Policy;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -30,6 +31,7 @@ import java.util.UUID;
  */
 public class FullTrigger {
 
+  private Map<String,String> actionToTriggerActionMap;
 
   public Trigger trigger;
   public List<Condition> conditions;
@@ -39,6 +41,9 @@ public class FullTrigger {
     trigger = new Trigger();
     conditions = new ArrayList<>();
     trigger.actions = new HashSet<>();
+    actionToTriggerActionMap = new HashMap<>();
+    actionToTriggerActionMap.put("webhook","hooks");
+    actionToTriggerActionMap.put("email","email");
   }
 
   public FullTrigger(Policy policy, boolean generatePseudoId) {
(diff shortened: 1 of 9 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/model/engine/FullTrigger.java
```diff
@@ -90,25 +95,52 @@ private static String generateId() {
...
+        actionName = actionIn;
+      }
+      actionName = actionToTriggerActionMap.get(actionName);
+
+      boolean found = false;
+      for (TriggerAction ta : trigger.actions) {
+        String taName = ta.actionPlugin;
+        if (taName.equals(actionName)) {
+          newActions.add(ta);
+          found = true;
```
josejulio: Could you add a `break` after this? 
I suppose that in the future we are also going to check for actionParams (as currently we have no params)

src/main/java/com/redhat/cloud/custompolicies/app/model/engine/FullTrigger.java
```diff
@@ -90,25 +95,52 @@ private static String generateId() {
...
+        }
+      }
+      if (!found) {
+        var ta = new TriggerAction();
+        ta.actionPlugin = actionName;
+        newActions.add(ta);
+      }
+    }
+    trigger.actions.removeIf(ta -> !newActions.contains(ta));
```
josejulio: Should we remove this line?
As I understand, `newActions` contains the actions that we want to keep and the new actions.

src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
```diff
@@ -403,19 +403,42 @@ public Response updatePolicy(@QueryParam ("dry") boolean dryRun, @PathParam("pol
...
-          Policy.persist(storedPolicy);
-          FullTrigger trigger = new FullTrigger(storedPolicy);
 
+          boolean okToStore;
+          existingTrigger.updateFromPolicy(storedPolicy);
           if (!skipEngineCall) {
             try {
-              engine.updateTrigger(storedPolicy.id, trigger, false, user.getAccount());
+              engine.updateTrigger(storedPolicy.id, existingTrigger, false, user.getAccount());
+              okToStore = true;
```
josejulio: Is this `okToStore` really needed? Maybe missed a case.
- If the engine failed, it would throw an exception and would return error response (thus not storing)
- If the engine succeed it would just persist
- If !skipEngine, it would just persist

src/main/java/com/redhat/cloud/custompolicies/app/model/engine/FullTrigger.java
pilhuhn: Yes. So we need to remove from `trigger.actions` those that are not longer wanted, which that line does. As alternative, one could do `trigger.actions = newActions;`

src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
pilhuhn: I think you are right.

src/main/java/com/redhat/cloud/custompolicies/app/model/engine/FullTrigger.java
```diff
@@ -39,6 +41,9 @@ public FullTrigger() {
     trigger = new Trigger();
     conditions = new ArrayList<>();
     trigger.actions = new HashSet<>();
+    actionToTriggerActionMap = new HashMap<>();
+    actionToTriggerActionMap.put("webhook","hooks");
+    actionToTriggerActionMap.put("email","email");
```
josejulio: One last request, could you please use this `actionToTrigger` map for `storeActions` method ? I see it is still using the old if-else.

burmanm: Isn't that "webhook" -> "hooks" dangerous? What happens when we start to use "webhook" (which will require parameters and is not compatible with the old one) ? Will it need "realwebhookpleasestandup", "webhook" ?

pilhuhn: Good idea.

PR #70, opened 2020-03-23
CPOL-102 Also provide policy create time.

pilhuhn:
- Fix a bug in mtime handling as well.

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
@@ -435,12 +435,11 @@ public class PolicyCrudService {
           storedPolicy.populateFrom(policy);
           storedPolicy.customerid = user.getAccount();
           storedPolicy.setMtimeToNow();
-          FullTrigger trigger = new FullTrigger(storedPolicy);
 
           existingTrigger.updateFromPolicy(storedPolicy);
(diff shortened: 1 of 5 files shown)


josejulio:
Looks good, but has conflicts now.

src/main/java/com/redhat/cloud/custompolicies/app/rest/PolicyCrudService.java
```diff
@@ -434,11 +434,14 @@ public Response updatePolicy(@QueryParam ("dry") boolean dryRun, @PathParam("pol
 
           storedPolicy.populateFrom(policy);
           storedPolicy.customerid = user.getAccount();
+          storedPolicy.setMtimeToNow();
+          FullTrigger trigger = new FullTrigger(storedPolicy);
 
           existingTrigger.updateFromPolicy(storedPolicy);
```
josejulio: Is this not used anymore? What is it for? 
The `trigger` is not using the `updateFromPolicy`, so i assume we can remove that code?

pilhuhn: Yes, I think the `trigger` is rebase fallout. Will look at it later.

pilhuhn:
@josejulio The `existingTrigger` is the one requested from the engine, which has the id fields set etc. We need to send this one back to the engine (in modified form).

PR #72, opened 2020-03-23
CPOL-146 Add validation for passed actions

pilhuhn:

### DIFF
--- a/src/main/java/com/redhat/cloud/custompolicies/app/model/validation/ActionValidator.java
+++ b/src/main/java/com/redhat/cloud/custompolicies/app/model/validation/ActionValidator.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


src/main/java/com/redhat/cloud/custompolicies/app/model/validation/ActionValidator.java
```diff
@@ -0,0 +1,51 @@
...
+      }
+      String[] actions = input.split(";");
+      int i = 0;
+      for (String action: actions) {
+         String a = action.toLowerCase().strip();
+         if (a.isEmpty() || validActions.contains(a)) {
+            i++;
+         }
+      }
+      return i == actions.length;
```
josejulio: I don't think we want to support multiple ways to specify an empty actions ( i.e. `";"`), the default `""` seems enough.
I'm curious if you had in mind something else.

pilhuhn: This is still one string that can list multiple actions. Not yet an array.
So it could be "email;webhook" or someone could supply "email ; "

pilhuhn:
close + reopen did not trigger a new build ;(

PR #74, opened 2020-03-24
POL-147 Rename project from Custom-Policies to Policies

josejulio:


josejulio:
Travis didn't run, not sure if we need to do something there

PR #78, opened 2020-03-26
POL-97 Sets lastEvaluation to empty when 0

josejulio:
Prevents having to parse stuff like "1969-12-31 18:00:00.0" as "not set"

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/model/Policy.java
+++ b/src/main/java/com/redhat/cloud/policies/app/model/Policy.java
@@ -90,7 +90,7 @@ public class Policy extends PanacheEntityBase {
   private Timestamp mtime=new Timestamp(System.currentTimeMillis());
 
   @Schema(type = SchemaType.STRING,
-          description = "Last evaluation time in a form like '2020-01-24 12:19:56.718', output only",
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/cloud/policies/app/model/Policy.java
```diff
@@ -120,7 +120,11 @@ public String getMtime() {
 
   @JsonbTransient
   public void setLastEvaluation(long lastEvaluation) {
-    this.lastEvaluation = new Timestamp(lastEvaluation);
+    if (lastEvaluation == 0) {
+      this.lastEvaluation = null;
+    } else {
+      this.lastEvaluation = new Timestamp(lastEvaluation);
```
burmanm: A small note - this class is incorrect. ``java.sql.Timestamp`` is most certainly not what one should use. Perhaps you wanted ``LocalDateTime`` instead? (not broken in this commit - this has been broken all along)

burmanm: Although I'm not sure why not just send the epoch ..

josejulio: I think we probably should use the epoch. Is easier to work with it and we are losing the timezone when convertir it to `Timestamp`.

PR #79, opened 2020-03-30
POL-169 Get rid of the '.0' version of the api naming

pilhuhn:


theute:
Let's see if we can get tools to adapt, rather than adapt to tools :-(
So let's not merge this now

PR #81, opened 2020-03-31
Bump Quarkus to 1.3.1

pilhuhn:

I have no idea if there is a need for a jdbc driver change in the DC when this is installed or not.
There is certainly a change in application.properties respective to 1.3.0 for the case of tests.


theute:
Pressing the merge button, praying and ready to fix if needed

theute:
Seems to work with no change to the datasource

PR #82, opened 2020-03-31
POL-182 Change return code of DELETE policy from 400 to 404

pilhuhn:


theute:
Travis passed, but didn't report back (and that's not the first time those days...)

PR #89, opened 2020-04-06
Supports NO_LIMIT (-1) queries

josejulio:


src/main/java/com/redhat/cloud/policies/app/model/Policy.java
```diff
@@ -173,7 +173,9 @@ public String getCtime() {
     // Always set the customer id - even if that may be part of parameters below
     q.setParameter("customerid",customer);
     q.setFirstResult(pager.getOffset());
```
pilhuhn, line 175: Should we start from page 1 in that case no matter on which the user stands?

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -53,23 +53,16 @@ public static Pager extractPager(@NotNull UriInfo uriInfo) {
...
-
         String itemsPerPage = queryParams.getFirst(QUERY_LIMIT);
+        boolean usingNoLimit = false;
         if (itemsPerPage != null) {
             try {
-                pageBuilder.itemsPerPage(Integer.parseInt(itemsPerPage));
+                int limit = Integer.parseInt(itemsPerPage);
+                if (limit == Pager.NO_LIMIT) {
+                    usingNoLimit = true;
+                    pageBuilder.page(0);
```
pilhuhn, line 63: Ah, you do it here. It may be still more clear to have the `setFirstResult` above within the conditional.

josejulio: MMh, it's true, here I'm assuming that the default is to ignore across the PagingUtils, as is more convenient for tests. I can replicate that check also in the policies in case we decide to change this later.

PR #91, opened 2020-04-09
POL-175 Support for mass-operations

pilhuhn:
- Provide policy ids after filtering.

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -24,6 +24,7 @@ import com.redhat.cloud.policies.app.model.Msg;
 import com.redhat.cloud.policies.app.model.Policy;
 import java.net.ConnectException;
 import java.net.URI;
+import java.util.List;
 import java.util.UUID;
 import java.util.logging.Logger;
 import javax.enterprise.context.RequestScoped;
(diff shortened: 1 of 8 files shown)


pilhuhn:
Will need rebasing once the others are merged, but is mostly done.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -240,6 +242,104 @@ public Response getPoliciesForCustomer() {
...
+                          },
+                          defaultValue = "equal"
+                  )
+          ),
+  })
+  @APIResponse(responseCode = "400", description = "Bad parameter for sorting was passed")
+  @APIResponse(responseCode = "404", description = "No policies found for customer")
+  @APIResponse(responseCode = "403", description = "Individual permissions missing to complete action")
+  @APIResponse(responseCode = "200", description = "PolicyIds found", content =
+                 @Content(schema = @Schema(implementation = List.class)),
```
josejulio, line 320: I think this should be a PagedResponse (looking at what is being returned), otherwise just return the List below.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -356,27 +456,72 @@ public Response deletePolicy(@PathParam("id") UUID policyId) {
...
+          engine.deleteTrigger(policy.id, user.getAccount());
+          deletedOnEngine = true;
+        } catch (NotFoundException nfe) {
+          // Engine does not have it - we can delete anyway
+          deletedOnEngine = true;
+        } catch (Exception e) {
+          log.warning("Deletion on engine failed because of " + e.getMessage());
+        }
+        if (deletedOnEngine) {
+          policy.delete();
```
josejulio, line 516: I'm not really sure about this operation, at this point, we have already deleted the policy from the engine, lets say it fails on the last (out of 10) policy (for whatever reason), then we won't delete the policy from the backend but it will be gone from the engine. I think we just should catch this exception and skip the `deleted.add(uuid)`

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
pilhuhn: I am not sure I follow. You mean to return a paged list to be consistent with getPolicies?

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
pilhuhn: If it fails for whatever reason to delete on the engine, `deletedOnEngine` is false and `policy.delete()` would not kick in.
Perhaps you mean something else?

josejulio: I mean if it fails in `policy.delete`, although maybe not that likely.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
josejulio: No. I mean that it currently returns a PagedResponse (see below: `return PagingUtils.responseBuilder(page).build();
`). I think a `List` is better, as we don't paginate on this method, but that's not what this method is returning.

```
$  curl 'https://ci.foo.redhat.com:1337/api/policies/v1.0/policies/ids?offset=0&limit=50' # etc
{"data":["31695b91-9c45-4162-8aa5-b7ef5e9ea23b","d7073ec1-2e11-4353-91ad-06d5865b6c65","9f1b9a76-de92-4ee9-979e-b548b65c40d4"],"links":{"last":"/api/policies/v1.0/policies?limit=50&offset=0","first":"/api/policies/v1.0/policies?limit=50&offset=0"},"meta":{"count":3}}
```

pilhuhn: Now I get it. I think I was looking at the wrong place and thus misunderstood. Will change that later and also for enable/disable + delete to simple lists

josejulio:
Tested with https://github.com/RedHatInsights/policies-ui-frontend/pull/102 and looks good!

PR #92, opened 2020-04-14
POL-214 Don't store policy if engine reports an error.

pilhuhn:


src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -297,14 +302,13 @@ public Response storePolicy(@QueryParam ("alsoStore") boolean alsoStore, @Valid
...
-    UUID id;
+    UUID id = null;
     try {
-      id = policy.store(user.getAccount(), policy);
-      // We persisted locally, now tell the engine
       if (!skipEngineCall) {
         FullTrigger trigger = new FullTrigger(policy);
         try {
           engine.storeTrigger(trigger, false, user.getAccount());
+          id = policy.store(user.getAccount(), policy);
```
josejulio, line 311: But this won't store in `skipEngineCall` is `true`.

I personally no longer use it and is not that hard to setup and engine, should we remove that now?

pilhuhn: yes, I started that in a different branch and you are right. Will do that now

PR #93, opened 2020-04-14
POL-215 - fix empty path

pilhuhn:
Also add some general validation of the open api document.

Btw: technically an empty path element is valid: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#securityFiltering

### DIFF
--- a/src/test/java/com/redhat/cloud/policies/app/OapiTest.java
+++ b/src/test/java/com/redhat/cloud/policies/app/OapiTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/test/java/com/redhat/cloud/policies/app/OapiTest.java
```diff
@@ -0,0 +1,62 @@
...
+ 			}
+ 			Assert.fail("OpenAPI spec is not valid");
+ 		}
+ 		//
+ 		// Now that basic validation is done, we can add some of our own
+		//
+		Map<String, Path> paths = model.getPaths();
+
+ 		// The base path filler. See also OASModifier.mangleName
+ 		Assert.assertTrue(paths.containsKey("/"));
```
josejulio, line 60: If i understood properly, shouldn't we also assert that it does not contain the key `""` ?

pilhuhn: It won't. An empty path is allowed according to spec to sort of hide things. Without the change in mangleName(), the / key will just not show up.

PR #96, opened 2020-04-17
POL-226 Retrieve lastTriggered time from engine.

pilhuhn:

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -247,13 +247,12 @@ public class PolicyCrudService {
         Map<String, Long> idTimeMap = new HashMap<>(page.size());
         for (Trigger fullTrigger : list) {
           String tid = fullTrigger.id;
-          boolean found = false;
           for (Map<String, Object> map : fullTrigger.lifecycle) {
-            if (!found && map.containsKey("status") && map.get("status").equals("ALERT_GENERATE")) {
+            if (map.containsKey("status") && map.get("status").equals("ALERT_GENERATE")) {
               BigDecimal aTime = (BigDecimal) map.get("stime");
               long tse =  aTime.longValue();
               idTimeMap.put(tid,tse);
-              found = true;
+              break;
             }
           }
         }
@@ -800,14 +799,12 @@ public class PolicyCrudService {
       try {
         List<Trigger> alerts = engine.findTriggersById(policyId.toString(), user.getAccount());
         if (alerts != null && !alerts.isEmpty()) {
-          System.out.println(alerts.get(0).lifecycle);
-          boolean found = false;
           for (Map<String, Object> map : alerts.get(0).lifecycle) {
-            if (!found && map.get("status").equals("ALERT_GENERATE")) {
+            if (map.containsKey("status") && map.get("status").equals("ALERT_GENERATE")) {
               BigDecimal aTime = (BigDecimal) map.get("stime");
               long tse =  aTime.longValue();
               policy.setLastTriggered(tse);
-              found = true;
+              break;
             }
           }
         }
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -245,16 +245,9 @@ public class PolicyCrudService {
 
         // Put the lastTriggered time (= creation time ) of the alert in a triggerId-alert map
         Map<String, Long> idTimeMap = new HashMap<>(page.size());
-        for (Trigger fullTrigger : list) {
-          String tid = fullTrigger.id;
-          for (Map<String, Object> map : fullTrigger.lifecycle) {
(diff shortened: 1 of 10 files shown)


josejulio:
Haven't tested, but code looks good, i suppose this one is waiting for the changes on the engine?

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -235,6 +226,35 @@ public Response getPoliciesForCustomer() {
...
+            if (stored < a.ctime) {
+              times.put(tid, a.ctime);
+            }
+          }
+        }
+        // Now loop over the map and transfer the times to the policies
+        page.forEach(p -> {
+          if (times.containsKey(p.id.toString())) {
+            long lastTriggerTime = times.get(p.id.toString());
+            p.setLastTriggered(lastTriggerTime);
```
josejulio, line 250: unix timestamp?

pilhuhn: Yes.
I can also convert that to Timestamp. But as we had the discussion that Unix ts may be better, I did it this way.

josejulio: Unix timestamp is good, thanks!

pilhuhn:
@josejulio It looks good locally. Was not able to locally test it with ui-pr 121 though

josejulio:
> @josejulio It looks good locally. Was not able to locally test it with ui-pr 121 though

Great! I'll test it out.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -232,18 +237,36 @@ public Response getPoliciesForCustomer() {
...
-      page.forEach(p -> {
-        try {
-          FullTrigger ft = engine.fetchTrigger(p.id, user.getAccount());
-          if (ft.conditions != null && !ft.conditions.isEmpty()) {
-            p.setLastEvaluation(ft.conditions.get(0).lastEvaluation);
+
+      // Get lastTriggeredTime and add it to the policies
+      String idsToFetch = page.stream().map(p -> p.id.toString()).collect(Collectors.joining(","));
+      try {
+        List<Trigger> list = engine.findTriggersById(idsToFetch,   user.getAccount());
```
josejulio, line 244: When the engine throws an error 500, this makes this request to fail too, not sure if that's expected or desired.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -232,18 +237,36 @@ public Response getPoliciesForCustomer() {
...
+        Map<String, Long> idTimeMap = new HashMap<>(page.size());
+        for (Trigger fullTrigger : list) {
+          String tid = fullTrigger.id;
+          boolean found = false;
+          for (Map<String, Object> map : fullTrigger.lifecycle) {
+            if (!found && map.containsKey("status") && map.get("status").equals("ALERT_GENERATE")) {
+              BigDecimal aTime = (BigDecimal) map.get("stime");
+              long tse =  aTime.longValue();
+              idTimeMap.put(tid,tse);
+              found = true;
```
josejulio, line 256: I think it would be better to remove the `found` and just add a `break` here.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -775,12 +798,21 @@ public Response getPolicy(@PathParam("id") UUID policyId) {
       builder = Response.status(Response.Status.NOT_FOUND);
     } else {
       try {
-        FullTrigger ft = engine.fetchTrigger(policyId, user.getAccount());
-        if (ft.conditions != null && !ft.conditions.isEmpty()) {
-          policy.setLastEvaluation(ft.conditions.get(0).lastEvaluation);
+        List<Trigger> alerts = engine.findTriggersById(policyId.toString(), user.getAccount());
+        if (alerts != null && !alerts.isEmpty()) {
+          System.out.println(alerts.get(0).lifecycle);
```
josejulio, line 803: I suppose you forgot to remove this one?

josejulio:
I tested this one with the UI code.
I'm not sure how I could make some policies to trigger manually to test this one better, but I saw that I was getting this new lastTriggered property on the returned json.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
pilhuhn: I don't get the comment.
If the engine call in line 244 fails, then lastTriggered time will be 0.

josejulio: I had a bad/old engine running (fixed with a `mvn clean install`). All the requests failed here.


The `} catch (Exception e) {  p.setLastEvaluation(0);  }` clause is removed and the `ProcessingException|NotFoundException pe`  does not catch this error.

PR #99, opened 2020-04-22
Introduce explicit timeouts for calls to other services.

pilhuhn:
This way the user on the UI does not need to wait for too long if they are in a odd state.

Before this change if you Ctrl-Z the engine, and then refresh the list or click on store in the create-wizard, it would hang "forever". With this change, the rest-request to engine, rbac and notifications is aborted after ~2s and the user either gets a nice error message or in case of list view, the list is shown, but data like lastEvaluated (lastTriggered in the future) will be omitted.

cc @hsagman -- this is something we talked about yesterday

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -34,16 +34,7 @@ import javax.persistence.EntityManager;
 import javax.persistence.PersistenceException;
 import javax.transaction.Transactional;
 import javax.validation.Valid;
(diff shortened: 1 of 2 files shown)


josejulio:
> Before this change if you Ctrl-Z the engine, and then refresh the list or click on store in the create-wizard, it would hang "forever". With this change, the rest-request to engine, rbac and notifications is aborted after ~2s and the user either gets a nice error message or in case of list view, the list is shown, but data like lastEvaluated (lastTriggered in the future) will be omitted.

This was hard to test for me, i think I was able to see it once.

Anyway i didn't see anything breaking.

PR #101, opened 2020-04-24
Do some Coverage recording and reporting

pilhuhn:
This is WIP for now, as JaCoCo sometimes throws strange errors when writing the .exec file.
Also unit test data seems not yet recorded. But then I am not sure it matters too much.

Should we move the coverage recording into a special profile, that is only executed on demand?


pilhuhn:
Latest coverage data is bad because of https://travis-ci.org/github/RedHatInsights/policies-ui-backend/builds/679012790#L517 -- that error shows up from time to time (but not always). See also https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/JaCoCo.20coverage.20and.20JDK.20versions

pilhuhn:
Bildschirmfoto 2020-04-24 um 17 25 54

Current status if the Jacoco issue does not hit.

josejulio:
> Current status if the Jacoco issue does not hit.

Is there any github issue that we should follow?

pilhuhn:
No issue (yet). Rather this discussion: https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/JaCoCo.20coverage.20and.20JDK.20versions

PR #103, opened 2020-04-27
PoC: Sync down policies into the engine.

pilhuhn:
PoC of a service, that gets all Policies from the DB, checks if they exist in the engine and if not,
stores them in the engine.

When the app starts, a token is emitted in the logs. This token needs to be given on the URL for the mechanism to work. Also it currently requires a valid rh-identity.
Currently it does not try to sync entries that are different in postgres and engine.

Ideally we would be able to block the endpoint in 3scale.

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/JaxRsApplication.java
+++ b/src/main/java/com/redhat/cloud/policies/app/JaxRsApplication.java
@@ -47,6 +47,8 @@ public class JaxRsApplication extends Application {
     router.route().order(-1000).handler(handler);
     showVersionInfo();
 
+    // Generate a token
+    TokenHolder.getInstance();
   }
 
   private void showVersionInfo() {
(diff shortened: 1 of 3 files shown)


burmanm:
What's the point?

theute:
I guess this is what I mentioned earlier today, in prod we can't offer to lose all the data of the engine.
Luckily we can recreate the triggers with the data that the users gave us (still losing the history).

burmanm:
This is no substitute for backups.

josejulio:
Code looks OK to me. 
This implies (to me) that we don't really trust the engine database much, should we do something about it?

pilhuhn:
Usage: 
```
 curl -XPOST -i -Hx-rh-identity:`cat server/src/test/resources/rhid.txt` http://localhost:8080/api/policies/v1.0/policies/sync?token=<token>
```

PR #106, opened 2020-04-30
POL-241 Validate policy name

josejulio:


src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -711,6 +711,31 @@ public Response validateCondition(Policy policy) {
...
 
+  @Operation(summary = "Validates a Policy condition")
+  @POST
+  @Path("/validate-name")
+  @APIResponses({
+          @APIResponse(responseCode = "200", description = "Name validated"),
+          @APIResponse(responseCode = "400", description = "Name not valid"),
+          @APIResponse(responseCode = "500", description = "No policy provided or internal error")
+  })
+  public Response validateName(Policy policy) {
```
pilhuhn, line 722: Policy should be `@Valid` and (see my other PR) `@NotNull`. 
The latter also removes the need for the null check a few lines down.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -711,6 +711,31 @@ public Response validateCondition(Policy policy) {
...
+    if (!user.canReadAll()) {
+      return Response.status(Response.Status.FORBIDDEN).entity(new Msg("Missing permissions to verify policy")).build();
+    }
+
+    if (policy == null) {
+      return Response.status(500, "No policy passed").build();
+    }
+
+    Response isNameValid = isNameValid(policy);
```
pilhuhn, line 731: 409 response code is not documented

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -626,9 +626,9 @@ public Response updatePolicy(@QueryParam ("dry") boolean dryRun, @PathParam("pol
         builder = Response.status(400, "Invalid policy");
       } else {
 
-        Policy tmp = Policy.findByName(user.getAccount(), policy.name);
-        if (tmp != null && !tmp.id.equals(policy.id)) {
-          return Response.status(409).entity(new Msg("Policy name is not unique")).build();
+        Response invalidNameResponse = isNameValid(policy);
+        if (invalidNameResponse != null) {
+          return invalidNameResponse;
```
pilhuhn, line 631: 409 code is not documented

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -747,4 +772,14 @@ public Response getPolicy(@PathParam("id") UUID policyId) {
     return builder.build();
   }
 
+  private Response isNameValid(Policy policy) {
```
pilhuhn, line 775: This should perhaps be renamed to `isNameUnique` or similar.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
josejulio: I think I'll skip the `@Valid` or i would have to provide a dummy condition for this call (but will use the `@NotNull`.

Any concern with that?

pilhuhn: I guess it is ok if we/you document this.
And then, if it is only about name validation, we could only pass a String that needs to fit the  rules for a name (not empty, max 150 chars) and not a Policy object.

josejulio: Ah, I suppose I could just pass the name. 

I wanted to keep the name validation, but without `@Valid` is not being used anyway.

josejulio: Ok, changed it to receive the name in the body, and the Id (in case is an existing policy) with query param.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -706,6 +717,42 @@ public Response validateCondition(@Valid @NotNull Policy policy) {
...
+  @POST
+  @Path("/validate-name")
+  @APIResponses({
+          @APIResponse(responseCode = "200", description = "Name validated"),
+          @APIResponse(responseCode = "400", description = "Policy validation failed"),
+          @APIResponse(responseCode = "409", description = "Name not unique"),
+          @APIResponse(responseCode = "500", description = "Internal error")
+  })
+  public Response validateName(@NotNull String policyName, @QueryParam("id") UUID id) {
+    if (!user.canReadAll()) {
```
pilhuhn, line 730: Code 403 not in openapi

josejulio: Added

PR #108, opened 2020-05-04
POL-207 Limit Policy.name to 150 characters

josejulio:
Also guards the API calls for adding or editing policies with long names.


pilhuhn:
This is strange. I am pretty sure I did that a while ago

josejulio:
> @josejulio did you check that OpenAPI reflects the `@Size` parameter?

Yes, it does reflect on OpenAPI. `@Length` does not =-(.

With `@Size`:

Screenshot_2020-05-05_08-40-29

pilhuhn:
I saw that `@Length` did not reflect it, so I put it somewhere else :)

PR #109, opened 2020-05-11
POL-270 Filter for enabled column is always BOOLEAN_IS

pilhuhn:

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
@@ -118,10 +118,19 @@ public class PagingUtils {
             if (filterMatcher.find()) {
                 String column = filterMatcher.group(1);
                 String value = queryParams.getFirst(key);
-                String operatorString = queryParams.getFirst(String.format("%s[%s]", FILTER_OP, column));
-                Filter.Operator operator = Filter.Operator.EQUAL;
-                if (operatorString != null) {
-                    operator = Filter.Operator.fromName(operatorString);
+                Filter.Operator operator;
+                if (column.equals("is_enabled")) {
+                    operator = Filter.Operator.BOOLEAN_IS;
+                    if (value==null || value.isEmpty()) {
+                        value = "true";
+                    }
+                }
+                else {
+                    String operatorString = queryParams.getFirst(String.format("%s[%s]", FILTER_OP, column));
+                    operator = Filter.Operator.EQUAL;
+                    if (operatorString != null) {
+                        operator = Filter.Operator.fromName(operatorString);
+                    }
                 }
                 pageBuilder.filter(column, operator, value);
             }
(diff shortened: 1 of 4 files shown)


pilhuhn:
@josejulio I wonder if we should remove the BOOLEAN_IS from the other columns.
And perhaps rewrite the entire Parameter list to only list filter[] and filter:op[] once, but with explanation and examples.

josejulio:
> @josejulio I wonder if we should remove the BOOLEAN_IS from the other columns.

This makes sense, no point in using that `BOOLEAN_IS` in other columns.

> And perhaps rewrite the entire Parameter list to only list filter[] and filter:op[] once, but with explanation and examples.

It would work, but is more explicit separating the columns and actually giving the specific parameters that are valid for each one instead of all-to-all.

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -118,10 +118,22 @@ public static Pager extractPager(@NotNull UriInfo uriInfo) {
...
                 String value = queryParams.getFirst(key);
-                String operatorString = queryParams.getFirst(String.format("%s[%s]", FILTER_OP, column));
-                Filter.Operator operator = Filter.Operator.EQUAL;
-                if (operatorString != null) {
-                    operator = Filter.Operator.fromName(operatorString);
+                Filter.Operator operator;
+                if (column.equals("is_enabled")) {
+                    operator = Filter.Operator.BOOLEAN_IS;
+                    if (value==null || value.isEmpty()) {
+                        value = "true";
```
josejulio, line 125: +1

We could document that this value defaults to true when not found or empty:` @Parameter(name= filter[is_enabled]`

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -118,10 +118,22 @@ public static Pager extractPager(@NotNull UriInfo uriInfo) {
             if (filterMatcher.find()) {
                 String column = filterMatcher.group(1);
                 String value = queryParams.getFirst(key);
-                String operatorString = queryParams.getFirst(String.format("%s[%s]", FILTER_OP, column));
-                Filter.Operator operator = Filter.Operator.EQUAL;
-                if (operatorString != null) {
-                    operator = Filter.Operator.fromName(operatorString);
+                Filter.Operator operator;
+                if (column.equals("is_enabled")) {
```
josejulio, line 122: The only "problem" is that `PagingUtils` was meant to be general across all our resources. 
I think we will figure out what to do IF we want to have more resources filtered, not an issue right now.

pilhuhn: Yes. I had the same thoughts. We would need some datastructure (additional fields on the `Filter.Operator` enum) to list applicable fields or vice versa.
But yes, let's think about it when we need it.

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
pilhuhn: I wonder if the @Param for filter:op[is_enabled] should just go away

josejulio: Ah, for this case it makes sense to just remove it, is the only possible value.

PR #110, opened 2020-05-13
POL-241 Ensure validate-name decodes string from json

josejulio:
The frontend is encoding the content as a json, so `ABC` becomes `"ABC"`, but the backend is not honoring the `Content-Type` and taking the plain json string.

This fixes that call and makes the method to specifically expect a `JsonString`.


josejulio:
Checking how this affects the openapi.json.

PR #112, opened 2020-05-13
POL-276 Move to github actions and upload openapi.json

josejulio:


josejulio:
@pilhuhn I'm moving to github actions to be able to store an artifact as part of the run (in this case, the `openapi.json`). 
This file is attached to the run (see the [action](https://github.com/RedHatInsights/policies-ui-backend/pull/112/checks?check_run_id=671735070)) and can be be queried to be retrieved later.

The plan is to have the frontend to download latest artifact from master and check if is still valid similar to what is [currently being done](https://github.com/RedHatInsights/policies-ui-frontend/pull/142) with the `Expression.g4` file.

src/test/java/com/redhat/cloud/policies/app/OapiTest.java
```diff
@@ -64,5 +69,8 @@ public void validateOpenApi() throws Exception {
  		// User config is private, so don't show it
  		Assert.assertFalse(paths.containsKey("/user-config"));
  		Assert.assertFalse(schemas.containsKey("SettingsValues"));
+
+		InputStream in = url.openStream();
```
pilhuhn, line 73: I think it would be good to document this.
Someone new to the project may not understand why this exists in a test.

PR #114, opened 2020-05-21
Changes on policy paging

josejulio:
- Use range paging
- Fix a bug on last page when the total is divisible by the limit


josejulio:
Upgrading to quarkus 1.4.2 seems to break our openapi.json (or something in this PR), all the `filter:op[*]` are no longer there.

pilhuhn:
I see the missing `filter*` ones in a local branch on 1.4.2 as well. Only the `filter[is_enabled]` is present in openapi.json

josejulio:
Rebased (using quarkus 1.5.1)

PR #119, opened 2020-06-09
POL-293 Add paging to the history of a single policy

pilhuhn:

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -860,9 +860,9 @@ public class PolicyCrudService {
          for (Map<String,Object> value : list) {
            long ctime = (long) value.get("ctime");
            Map<String,Object> tmp = (Map<String, Object>) value.get("tags");
-           String insights_id = (String) tmp.get("inventory_id");
+           String inventory_id = (String) tmp.get("inventory_id");
            String name = (String) tmp.get("display_name");
-           HistoryItem hi = new HistoryItem(ctime,insights_id,name);
+           HistoryItem hi = new HistoryItem(ctime,inventory_id,name);
            items.add(hi);
          }
          Page<HistoryItem> itemsPage = new Page<>(items,pager,totalCount);
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -83,6 +83,8 @@ import org.eclipse.microprofile.openapi.annotations.responses.APIResponses;
 import org.eclipse.microprofile.rest.client.inject.RestClient;
 import org.hibernate.exception.ConstraintViolationException;
 
+import static java.lang.Integer.min;
+
 /**
  * @author hrupp
  */
@@ -128,7 +130,7 @@ public class PolicyCrudService {
           @Parameter(
                   name = "limit",
                   in = ParameterIn.QUERY,
-                  description = "Number of items per page, if not specified uses 10. " + Pager.NO_LIMIT + " can be used to specify an unlimited page, when specified it ignores the offset" ,
+                  description = "Number of items per page, if not specified uses 50. " + Pager.NO_LIMIT + " can be used to specify an unlimited page, when specified it ignores the offset" ,
                   schema = @Schema(type = SchemaType.INTEGER)
           ),
           @Parameter(
@@ -786,12 +788,26 @@ public class PolicyCrudService {
   @APIResponse(responseCode = "403", description = "Individual permissions missing to complete action")
   @APIResponse(responseCode = "404", description = "Policy not found")
   @APIResponse(responseCode = "500", description = "Retrieval of History failed")
-  @Parameter(name = "id", description = "UUID of the policy")
+  @Parameters({
+      @Parameter(
+          name = "offset",
+          in = ParameterIn.QUERY,
+          description = "Page number, starts 0, if not specified uses 0.",
+          schema = @Schema(type = SchemaType.INTEGER)
+      ),
+      @Parameter(
+          name = "limit",
+          in = ParameterIn.QUERY,
+          description = "Number of items per page, if not specified uses 50. Maximum value is 200.",
+          schema = @Schema(type = SchemaType.INTEGER)
+      ),
+      @Parameter(name = "id", description = "UUID of the policy")
+  })
   @GET
   @Path("/{id}/history/trigger")
   public Response getTriggerHistoryForPolicy(@PathParam("id") UUID policyId) {
     if (!user.canReadAll()) {
-       return Response.status(Response.Status.FORBIDDEN).entity(new Msg("Missing permissions to retrieve policies")).build();
+       return Response.status(Response.Status.FORBIDDEN).entity(new Msg("Missing permissions to retrieve the policy history")).build();
      }
 
      Policy policy = Policy.findById(user.getAccount(), policyId);
@@ -800,8 +816,16 @@ public class PolicyCrudService {
      if (policy==null) {
        builder = Response.status(Response.Status.NOT_FOUND);
      } else {
+
        try {
-         String alerts = engine.findLastTriggered(policyId.toString(), false, user.getAccount());
+         Pager pager = PagingUtils.extractPager(uriInfo);
+
+         int limit = pager.getLimit();
+         limit = min(limit,200);
+         int pageNum = pager.getOffset() / limit;
+
+         String alerts = engine.findLastTriggered(policyId.toString(), false, pageNum, limit,
+             user.getAccount());
 
          List<HistoryItem> items = new ArrayList<>();
           DocumentContext jp = JsonPath.parse(alerts);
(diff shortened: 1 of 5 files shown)


src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -800,8 +816,16 @@ public Response getTriggerHistoryForPolicy(@PathParam("id") UUID policyId) {
...
+
        try {
-         String alerts = engine.findLastTriggered(policyId.toString(), false, user.getAccount());
+         Pager pager = PagingUtils.extractPager(uriInfo);
+
+         int limit = pager.getLimit();
+         limit = min(limit,200);
+         int pageNum = pager.getOffset() / limit;
+
+         String alerts = engine.findLastTriggered(policyId.toString(), false, pageNum, limit,
```
josejulio, line 827: I'll need a "paginated response", e.g. to know the total number of elements to decide if "next page" button is disabled or not.

josejulio:
Something else to note, is that the UI would still need all the elements in the page for exporting (json/csv), should the UI consume all the pages to build this report?

josejulio:
Some test, not sure if i'm wrongly fetching the second page

```bash
curl 'https://ci.foo.redhat.com:1337/api/policies/v1.0/policies/3f421b5c-babc-4cf4-9026-0b9083d5b1df/history/trigger?limit=50&offset=0' | jq ". | length"
50
``` 

```bash
curl 'https://ci.foo.redhat.com:1337/api/policies/v1.0/policies/3f421b5c-babc-4cf4-9026-0b9083d5b1df/history/trigger?limit=200&offset=0' | jq ". | length"
200
``` 

but
```bash
curl 'https://ci.foo.redhat.com:1337/api/policies/v1.0/policies/3f421b5c-babc-4cf4-9026-0b9083d5b1df/history/trigger?limit=50&offset=50' | jq ". | length"
0
```

josejulio:
@burmanm maybe a bug from the engine?
I see that the following arguments are being sent to the engine:

```java
@GET
  @Consumes("application/json")
  String findLastTriggered(@QueryParam("triggerIds") String triggerIds,
                                @QueryParam("thin") boolean thin,
                                @QueryParam("page") int page,
                                @QueryParam("per_page") int per_page,
                                @HeaderParam("Hawkular-Tenant") String customerId);
```

```java
engine.findLastTriggered("the-id", false, 1, 50, "the-account")
```
and the engine returns `[]`

josejulio:
I updated the engine, and now I'm getting `[]` always, but I can see "Last Triggered" date,maybe related to https://github.com/RedHatInsights/policies-engine/pull/80 ?

burmanm:
I don't see that ``[]`` in the master for all queries. Thin alerts shouldn't make a difference to this, but I'll have to investigate the apparent bug (maybe there's a double filtering done - so Infinispan + code and that causes the offsets to go wrong).

josejulio:
Debugging the backend code, it looks like something changed on the engine, it is returning (for first page of 2 elements):
```json
[
   {
      "eventType":"ALERT",
      "tenantId":"940527",
      "id":"fe5a82ed-129c-4ce0-8e3b-63ab4aac045e-1591732136644-2243f530-f544-410b-bb4c-3d6090f65cc4",
      "ctime":1591732136644,
      "dataSource":"_none_",
      "dataId":"fe5a82ed-129c-4ce0-8e3b-63ab4aac045e",
      "category":"ALERT",
      "text":"Fail if we are running on a x86_64 - We should be using arm.",
      "trigger":{
         "tenantId":"940527",
         "id":"fe5a82ed-129c-4ce0-8e3b-63ab4aac045e",
         "name":"Copy of Not arch x86_64",
         "description":"Fail if we are running on a x86_64 - We should be using arm.",
         "type":"STANDARD",
         "eventType":"ALERT",
         "eventCategory":null,
         "eventText":null,
         "severity":"MEDIUM",
         "actions":[
            {
               "tenantId":"940527",
               "actionPlugin":"email",
               "actionId":"_managed-instance-email-35e48b6487d9568e"
            }
         ],
         "autoDisable":false,
         "autoEnable":false,
         "autoResolve":false,
         "autoResolveAlerts":false,
         "autoResolveMatch":"ALL",
         "enabled":true,
         "firingMatch":"ALL",
         "source":"_none_"
      },
      "severity":"MEDIUM",
      "status":"OPEN",
      "lifecycle":[
         {
            "status":"OPEN",
            "stime":1591732136644
         }
      ]
   },
   {
      "eventType":"ALERT",
      "tenantId":"940527",
      "id":"fe5a82ed-129c-4ce0-8e3b-63ab4aac045e-1591732136583-20d4a793-fa41-4321-9f6d-006e4d062cfc",
      "ctime":1591732136583,
      "dataSource":"_none_",
      "dataId":"fe5a82ed-129c-4ce0-8e3b-63ab4aac045e",
      "category":"ALERT",
      "text":"Fail if we are running on a x86_64 - We should be using arm.",
      "trigger":{
         "tenantId":"940527",
         "id":"fe5a82ed-129c-4ce0-8e3b-63ab4aac045e",
         "name":"Copy of Not arch x86_64",
         "description":"Fail if we are running on a x86_64 - We should be using arm.",
         "type":"STANDARD",
         "eventType":"ALERT",
         "eventCategory":null,
         "eventText":null,
         "severity":"MEDIUM",
         "actions":[
            {
               "tenantId":"940527",
               "actionPlugin":"email",
               "actionId":"_managed-instance-email-35e48b6487d9568e"
            }
         ],
         "autoDisable":false,
         "autoEnable":false,
         "autoResolve":false,
         "autoResolveAlerts":false,
         "autoResolveMatch":"ALL",
         "enabled":true,
         "firingMatch":"ALL",
         "source":"_none_"
      },
      "severity":"MEDIUM",
      "status":"OPEN",
      "lifecycle":[
         {
            "status":"OPEN",
            "stime":1591732136583
         }
      ]
   }
]
```

and the backend query `"$.[*].evalSets..value"` is the one returning `[]` after update.

burmanm:
Well, the backend is now removing evalsets from alerts storage if that is set (in the application.properties), so searching them would be impossible.

josejulio:
Ok, when setting `engine.backend.ispn.alerts-thin` to `false` it returns the evalSet as did before.

josejulio:
What i said above (https://github.com/RedHatInsights/policies-ui-backend/pull/119#issuecomment-641486265) is still an issue. 

Using pages of 50 elements:
- If I have 99 elements, page 1 and page 2 show correctly. 
- When I have 100 elements, the page 2 yields no elements.
- Having 102 elements page 1 shows OK, page 2 empty and page 3 shows 2 elements.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
pilhuhn: Yes. I am not sure yet how to get that from the engine

pilhuhn:
> Something else to note, is that the UI would still need all the elements in the page for exporting (json/csv), should the UI consume all the pages to build this report?

I think the UI would need to iterate over the pages. Sending with unlimited could be much too much data that is dealt with in one go.

burmanm:
> Something else to note, is that the UI would still need all the elements in the page for exporting (json/csv), should the UI consume all the pages to build this report?

If there's export functionality, then it should probably just stream the data from engine? No point keeping any of this in memory elsewhere.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -782,26 +784,51 @@ public Response getPolicy(@PathParam("id") UUID policyId) {
 
   @Operation(summary = "Retrieve the trigger history of a single policy")
   @APIResponse(responseCode = "200", description = "History could be retrieved",
-      content = @Content (schema = @Schema(type = SchemaType.ARRAY, implementation = HistoryItem.class)))
+      content = @Content (schema = @Schema(implementation = PagingUtils.PagedResponse.class)),
```
josejulio, line 787: I have a problem here, using this syntax removes the `HistoryItem` from the openapi.json plus the `PagedResponse` doesn't have any info about its contents.

I found the following [1] workaround:
```java
// workaround for returning generic types: https://github.com/swagger-api/swagger-core/issues/498#issuecomment-74510379
  // This class is used only for swagger return type
  private static class PagedResponseOfHistoryItem extends PagingUtils.PagedResponse<HistoryItem> {
    private List<HistoryItem> data;
    private PagedResponseOfHistoryItem() {}
  }

  @Operation(summary = "Retrieve the trigger history of a single policy")
  @APIResponse(responseCode = "200", description = "History could be retrieved",
      content = @Content (schema = @Schema(implementation = PagedResponseOfHistoryItem.class)),
```

WDYT?

[1] https://github.com/swagger-api/swagger-core/issues/498#issuecomment-74510379

pilhuhn: I was thinking about something similar, as I don't like the type-erasure. 
Same for `getPolicies`

pilhuhn:
This depends now on https://github.com/RedHatInsights/policies-engine/pull/82

josejulio:
Tested this PR, is working fine.
- Openapi has the types for the paged policy and paged triggers.
- Paginated response is good

Already iterating over the pages when exporting.

PR #121, opened 2020-07-09
POL-311 Harden x-rh-identity evaluation

pilhuhn:
We now check for empty account inside the header and reject calls accordingly.


pilhuhn:
cc @ lahavyuv86

PR #125, opened 2020-07-21
Expose metrics from /proc/self/status via MP-Metrics.

pilhuhn:
Expose some of the metrics that the OS sees via MP-Metrics

Bildschirmfoto 2020-07-21 um 15 04 36


theute:
<3 Metrics LGTM

PR #127, opened 2020-07-22
POL-301 policy history filter

pilhuhn:
Now that POL-314 is in the engine, filtering on the history can be done.

### DIFF
--- a/src/test/java/com/redhat/cloud/policies/app/RestApiTest.java
+++ b/src/test/java/com/redhat/cloud/policies/app/RestApiTest.java
@@ -631,7 +631,7 @@ class RestApiTest extends AbstractITest {
   }
 
   @Test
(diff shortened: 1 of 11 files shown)


src/test/java/com/redhat/cloud/policies/app/RestApiTest.java
```diff
@@ -41,6 +41,7 @@
 import javax.json.bind.JsonbBuilder;
 
 import io.restassured.response.Response;
+import org.jetbrains.annotations.NotNull;
```
josejulio, line 44: `jetbrains` ?
Wondering if you are using the desired annotation.

pilhuhn: Nope.

PR #135, opened 2020-07-24
[POL-302] Implement Policy History sorting.

pilhuhn:


src/main/java/com/redhat/cloud/policies/app/PolicyEngine.java
```diff
@@ -126,6 +126,12 @@ Response findLastTriggered(@QueryParam("triggerIds") String triggerIds,
                                 @QueryParam("thin") boolean thin,
                                 @QueryParam("page") int page,
                                 @QueryParam("per_page") int per_page,
+                                // Once MP-RestClient 2.0 is avaliable, we can turn the next lines
+                                // into String[] and set the query param style accordingly.
+                                @QueryParam("sort") String sort1,
+                                @QueryParam("sort") String sort2,
+                                @QueryParam("order") String order1,
+                                @QueryParam("order") String order2,
```
josejulio, line 134: Is this for sorting by 2 columns?
I don't think the current UI components support that out out of the box.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -885,12 +911,34 @@ public Response getTriggerHistoryForPolicy(@PathParam("id") UUID policyId) {
...
+
+         if (pager.getSort()!= null) {
+           List<Sort.Column> columns = pager.getSort().getColumns();
+           if (columns.size() > 0) {
+             sort1 = getSortFromPageColumn(columns.get(0));
+             order1 = getOrderFromPageColumn(columns.get(0));
+           }
+           if (columns.size() > 1) {
+             sort2 = getSortFromPageColumn(columns.get(0));
+             order2 = getOrderFromPageColumn(columns.get(0));
```
josejulio, line 933: I suppose you mean the second element
```suggestion
             sort2 = getSortFromPageColumn(columns.get(1));
             order2 = getOrderFromPageColumn(columns.get(1));
```

src/main/java/com/redhat/cloud/policies/app/PolicyEngine.java
josejulio: Ah, but i see that it could sort using `ctime` by default on the second column.

pilhuhn: If we can only sort by one column, I am happy too :)

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
pilhuhn: Yeah

PR #138, opened 2020-07-30
POL-330 Honor sort order if no column is explicitly given.

pilhuhn:
This applies to the list of policies and the policy history

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
@@ -93,23 +93,22 @@ public class PagingUtils {
                 String column = columns.get(i);
                 Sort.Direction direction = Sort.Direction.Ascending;
                 if (directions != null && i < directions.size()) {
-                    switch(directions.get(i).toLowerCase()) {
-                        case "asc":
-                            direction = Sort.Direction.Ascending;
-                            break;
-                        case "desc":
-                            direction = Sort.Direction.Descending;
-                            break;
-                        default:
-                            throw new IllegalArgumentException("Unexpected sort order found: [" + columns.get(i) + "]");
-                    }
+                    String dir = directions.get(i);
+                    direction = getDirection(columns.get(i), dir);
                 }
                 pageBuilder.addSort(column, direction);
             }
         }
         else {
             // default sort is by mtime descending, so that newest end up on top
-            pageBuilder.addSort("mtime",Sort.Direction.Descending);
+            Sort.Direction direction = Sort.Direction.Descending;
+
+            // check if the user specified a sort order (but no column, so ctime is meant)
+            if (directions!=null && directions.size()>0) {
+                direction = getDirection("ctime", directions.get(0));
+            }
+
+            pageBuilder.addSort("mtime",direction);
         }
 
         for (String key: queryParams.keySet()) {
@@ -139,6 +138,21 @@ public class PagingUtils {
         }
 
         return pageBuilder.build();
+    }
+
+    private static Sort.Direction getDirection(String column, String dir) {
+        Sort.Direction direction;
+        switch(dir.toLowerCase()) {
+            case "asc":
+                direction = Sort.Direction.Ascending;
+                break;
+            case "desc":
+                direction = Sort.Direction.Descending;
+                break;
+            default:
+                throw new IllegalArgumentException("Unexpected sort order found: [" + column + "]");
+        }
+        return direction;
     }
 
     public static <T>ResponseBuilder responseBuilder(Page<T> page) {
--- a/src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
@@ -94,7 +94,7 @@ public class PagingUtils {
                 Sort.Direction direction = Sort.Direction.Ascending;
                 if (directions != null && i < directions.size()) {
                     String dir = directions.get(i);
-                    direction = getDirection(columns.get(i), dir);
+                    direction = getDirection(dir, columns.get(i));
                 }
                 pageBuilder.addSort(column, direction);
             }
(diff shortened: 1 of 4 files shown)


src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -93,23 +93,22 @@ public static Pager extractPager(UriInfo uriInfo) {
...
-                            throw new IllegalArgumentException("Unexpected sort order found: [" + columns.get(i) + "]");
-                    }
+                    String dir = directions.get(i);
+                    direction = getDirection(columns.get(i), dir);
                 }
                 pageBuilder.addSort(column, direction);
             }
         }
         else {
             // default sort is by mtime descending, so that newest end up on top
```
josejulio, line 103:
```suggestion
```

josejulio: ah nevermind, mtime is still added

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -93,23 +93,22 @@ public static Pager extractPager(UriInfo uriInfo) {
...
-            pageBuilder.addSort("mtime",Sort.Direction.Descending);
+            Sort.Direction direction = Sort.Direction.Descending;
+
+            // check if the user specified a sort order (but no column, so ctime is meant)
+            if (directions!=null && directions.size()>0) {
+                direction = getDirection("ctime", directions.get(0));
+            }
+
+            pageBuilder.addSort("mtime",direction);
```
josejulio, line 111: It's a bit odd that `ctime` is used by default when sort but no column is given and `mtime` is used by default when no column or sort is used.

I think that they both should be the same, probably `ctime`.

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -93,23 +93,22 @@ public static Pager extractPager(UriInfo uriInfo) {
...
             }
         }
         else {
             // default sort is by mtime descending, so that newest end up on top
-            pageBuilder.addSort("mtime",Sort.Direction.Descending);
+            Sort.Direction direction = Sort.Direction.Descending;
+
+            // check if the user specified a sort order (but no column, so ctime is meant)
+            if (directions!=null && directions.size()>0) {
+                direction = getDirection("ctime", directions.get(0));
```
josejulio, line 108: ah, this `ctime` is not actually honored, is a red herring.

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -93,23 +93,22 @@ public static Pager extractPager(UriInfo uriInfo) {
...
                 }
                 pageBuilder.addSort(column, direction);
             }
         }
         else {
             // default sort is by mtime descending, so that newest end up on top
-            pageBuilder.addSort("mtime",Sort.Direction.Descending);
+            Sort.Direction direction = Sort.Direction.Descending;
+
+            // check if the user specified a sort order (but no column, so ctime is meant)
```
pilhuhn, line 106: This should read mtime and not ctime

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
pilhuhn, line 108: Same here

pilhuhn:
I think I confused ctime and mtime because in the policy history it is ctime

src/main/java/com/redhat/cloud/policies/app/rest/utils/PagingUtils.java
```diff
@@ -93,23 +93,22 @@ public static Pager extractPager(UriInfo uriInfo) {
...
             // default sort is by mtime descending, so that newest end up on top
-            pageBuilder.addSort("mtime",Sort.Direction.Descending);
+            Sort.Direction direction = Sort.Direction.Descending;
+
+            // check if the user specified a sort order (but no column, so mtime is meant)
+            if (directions!=null && directions.size()>0) {
+                direction = getDirection(directions.get(0), "mtime");
+            }
+
+            pageBuilder.addSort("mtime",direction);
```
josejulio, line 111: The inconvenient with setting the sort order in this function is that we don't have `mtime` in the trigger history.

So (like the original issue), the contents of the following calls are the same:
- `curl 'https://ci.foo.redhat.com:1337/api/policies/v1.0/policies/c8d5cce5-7e04-4386-9875-8ccc0b27687b/history/trigger?limit=50&offset=0&sortOrder=desc'` 
- `curl 'https://ci.foo.redhat.com:1337/api/policies/v1.0/policies/c8d5cce5-7e04-4386-9875-8ccc0b27687b/history/trigger?limit=50&offset=0&sortOrder=asc'`
- `curl 'https://ci.foo.redhat.com:1337/api/policies/v1.0/policies/c8d5cce5-7e04-4386-9875-8ccc0b27687b/history/trigger?limit=50&offset=0'`

Perhaps we should pass some defaults to this function?

pilhuhn: I don't think that is an issue with the term 'mtime'. In https://github.com/RedHatInsights/policies-ui-backend/blob/9e9ad0d7fdbd7634045c591e051a4a42852c6cc5/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java#L978 it uses 'ctime' if the column name is not either 'id' or 'name'.

I think I know what you see: you use `sortOrder`- that is engine speak. The ui-backend uses `sortDirection` , which some un-named person introduced for getPolicies some time ago :-)

josejulio: My bad!

PR #144, opened 2020-08-18
Bump Quarkus to 1.7.3

pilhuhn:


josejulio:
When using this PR, the `Policy` in the `openapi.json` loses `ctime`, `mtime` and `lastTriggered`:

### Master
policy-master

### This PR
policy-pr-144

josejulio:
The issue goes away if we remove the `@JsonbTransient` from the setters, but some tests fail (UI seems fine).

```diff
diff --git a/src/main/java/com/redhat/cloud/policies/app/model/Policy.java b/src/main/java/com/redhat/cloud/policies/app/model/Policy.java
index 6866392..6420de7 100644
--- a/src/main/java/com/redhat/cloud/policies/app/model/Policy.java
+++ b/src/main/java/com/redhat/cloud/policies/app/model/Policy.java
@@ -98,7 +98,6 @@ public class Policy extends PanacheEntityBase {
   @Transient
   private long lastTriggered;
 
-  @JsonbTransient
   public void setMtime(String mtime) {
     this.mtime = Timestamp.valueOf(mtime);
   }
@@ -112,7 +111,6 @@ public class Policy extends PanacheEntityBase {
   }
 
 
-  @JsonbTransient
   public void setLastTriggered(long tTime) {
     lastTriggered = tTime;
   }
@@ -121,7 +119,6 @@ public class Policy extends PanacheEntityBase {
     return lastTriggered;
   }
 
-  @JsonbTransient
   public void setCtime(String ctime) {
     this.ctime = Timestamp.valueOf(ctime);
   }
```

Screenshot_2020-08-18_14-11-17

josejulio:
> When using this PR, the `Policy` in the `openapi.json` loses `ctime`, `mtime` and `lastTriggered`

Note that this only happens in the `openapi.json`, the returned data still includes these values.

pilhuhn:
WTF....
cc @maxandersen

maxandersen:
WTH :)

Is this smallrye or ?

pilhuhn:
Yes, that change seems to have come from here: https://github.com/quarkusio/quarkus/pull/11081

phillip-kruger:
This is SmallRye yes. I am looking at it to figure out where the change came from . See

https://github.com/smallrye/smallrye-open-api/issues/374
https://github.com/smallrye/smallrye-open-api/pull/384

Might be because it's on the setter, and not on the field. Not sure what the rule should be ? If it's on the setter, it should be ignore when creating a new object, but you should still be able to get the value when doing a GET ? However the model is the same for set and get ? So maybe we should ignore the ignore if not on a field ?

@MikeEdgar might also be able to help here

pilhuhn:
I recall that I had to 'play' with this for a while, as it has to work for OpenAPI export and JSON serialisation and Hibernate/JPA.

phillip-kruger:
Yea it's a problem when you want different behaviour for those. Maybe we should allow 'switching of' JsonB support using config in OpenAPI ?

MikeEdgar:
Yes, this certainly looks to have been caused by the tinkering with the "ignore" logic recently. How many models in this project are impacted?

phillip-kruger:
Logged a bug in SmallRye, @MikeEdgar will look at it and we will try and get it into 1.7.1

josejulio:
> Yes, this certainly looks to have been caused by the tinkering with the "ignore" logic recently. How many models in this project are impacted?

Only the Policy model has changes.

```diff
$ diff openapi.json  openapi-affected.json 
720,725d719
<           "ctime" : {
<             "format" : "yyyy-MM-dd hh:mm:ss.ddd",
<             "description" : "Create time in a form like '2020-01-24 12:19:56.718', output only",
<             "type" : "string",
<             "readOnly" : true
<           },
735,744d728
<           },
<           "lastTriggered" : {
<             "format" : "int64",
<             "type" : "integer"
<           },
<           "mtime" : {
<             "format" : "yyyy-MM-dd hh:mm:ss.ddd",
<             "description" : "Last update time in a form like '2020-01-24 12:19:56.718', output only",
<             "type" : "string",
<             "readOnly" : true

```

theute:
Looks like 1.7.3 just went out ;-)

pilhuhn:
On 9 Sep 2020, at 15:01, Thomas Heute wrote:

> Looks like 1.7.3 just went out ;-)

Aaaaaahhhrrggghhhh

:-)

PR #146, opened 2020-09-11
POL-287 Make the OpenAPI return type more specific than just 'list'

pilhuhn:
Josejulio can you check if that has any bad implications for the UI?
At least Swagger-UI is more happy now


josejulio:
I'll check if anything fails, but this change makes me more happy too.

josejulio (approved):
lgtm

PR #147, opened 2020-09-14
Change action to use webhook instead of old hooks

josejulio:


src/main/java/com/redhat/cloud/policies/app/model/engine/FullTrigger.java
```diff
@@ -41,7 +41,7 @@ public FullTrigger() {
     conditions = new ArrayList<>();
     trigger.actions = new HashSet<>();
     actionToTriggerActionMap = new HashMap<>();
-    actionToTriggerActionMap.put("webhook","hooks");
+    actionToTriggerActionMap.put("webhook","webhook");
```
pilhuhn, line 44: We could in theory omit the whole map now as both key and value are the same.

pilhuhn (approved):

src/main/java/com/redhat/cloud/policies/app/model/engine/FullTrigger.java
josejulio: Agree.

PR #151, opened 2020-09-18
Upgrading to Quarkus 1.8.1 and adding few more types

josejulio:
The quarkus upgrade removes from the openapi the EntityManager related fields.

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -273,11 +273,6 @@ public class PolicyCrudService {
     return PagingUtils.responseBuilder(page).build();
   }
 
-  private static class ResponseListOfUUIDS {
-    List<UUID> uuids;
-    private ResponseListOfUUIDS(List<UUID> uuids) { this.uuids=uuids;}
-  }
-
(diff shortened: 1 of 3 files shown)


pilhuhn (approved):

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -336,7 +331,7 @@ public Response getPoliciesForCustomer() {
   @APIResponse(responseCode = "404", description = "No policies found for customer")
   @APIResponse(responseCode = "403", description = "Individual permissions missing to complete action")
   @APIResponse(responseCode = "200", description = "PolicyIds found", content =
-                 @Content(schema = @Schema(implementation = ResponseListOfUUIDS.class)))
+                 @Content(schema = @Schema(type = SchemaType.ARRAY, implementation = UUID.class)))
```
pilhuhn, line 334: I think we have some other places like this.
Does this change not apply there as well?

josejulio: Probably, I'll check.

josejulio: Other places already use the `type = SchemaType.Array, implementation = UUID.class` stuff, and there is no other raw array that needs a change like this.

PR #154, opened 2020-09-25
POL-363 Don't sort the trigger history by id.

pilhuhn:
As previously discussed.
Question: should we silently do this. Or return some 400 for 'unknown sort field'?
The latter makes it probably more obvious for customers.


josejulio:
> Or return some 400 for 'unknown sort field'?

+1 I think we do that for policy filtering

pilhuhn:
Updated

josejulio (approved):
LGTM

PR #155, opened 2020-10-02
Split "MATCHES" filters and sets the value to lowercase

josejulio:
Matches filters stop working when there are spaces in the query, they don't
 match anything.
 One workaround is to split by spaces and use a separate query:
 e.g. tags.display_name LIKE "foo bar" =>
   ( tags.display_name matches *foo* and tags.display_name matches *bar* )

Setting it as draft, because not sure if this is what we really want, having spaces in this filter won't match the values in order:

e.g. `foo bar` will match `ETCfoo barETC` and `ABCfooETC bla barblabla`


josejulio:
I don't think we want to consider hosts that have spaces in the name. Which is what i was using to test. closing for now.

PR #157, opened 2020-10-07
Prototyping mass delete via JsonPatch

pilhuhn:
This will most likely never make it into the code base.
It is more to illustrate the idea.

The query-param on normal `deletePolicies()` is another idea on how to tackle the issue.

Related to POL-257


pilhuhn:
This is not going to happen

PR #159, opened 2020-10-08
Start of an admin endpoint at /admin.

pilhuhn:
Also move the sync endpoint to /admin.
Adds a live check that can be set to fail.

```
$ curl -i -X POST    http://localhost:8080/admin/status?status=degraded
{"msg":"Set degraded state"}
$curl http://localhost:8080/status
HTTP/1.1 500 Internal Server Error
Content-Length: 287
Content-Type: text/plain;charset=UTF-8

{
  engine=RESTEASY004655: Unable ...,
 admin-degraded=true,
 notifications=Unknown error, status code 401
}

$ curl -i  http://localhost:8080/health/live
HTTP/1.1 200 OK
content-type: application/json; charset=UTF-8
content-length: 121


{
    "status": "UP",
    "checks": [
        {
            "name": "live",
            "status": "UP"
        }
    ]
}

$ curl -i -X POST    http://localhost:8080/admin/status?status=admin-down
HTTP/1.1 200 OK
Content-Length: 30
Content-Type: application/json

{"msg":"Set admin down state"}

$curl http://localhost:8080/health/live
HTTP/1.1 503 Service Unavailable
content-type: application/json; charset=UTF-8
content-length: 201


{
    "status": "DOWN",
    "checks": [
        {
            "name": "live",
            "status": "DOWN",
            "data": {
                "status": "admin-down"
            }
        }
    ]
}

```

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/JaxRsApplication.java
+++ b/src/main/java/com/redhat/cloud/policies/app/JaxRsApplication.java
@@ -31,7 +31,7 @@ import java.util.logging.Logger;
  * @author hrupp
  *
  */
-@ApplicationPath("/api/policies/v1.0")
+@ApplicationPath("/")
 public class JaxRsApplication extends Application {
 
   Logger log = Logger.getLogger("Policies UI-Backend");
@@ -48,7 +48,7 @@ public class JaxRsApplication extends Application {
     showVersionInfo();
 
     // Generate a token
-    TokenHolder.getInstance();
+    StuffHolder.getInstance();
   }
 
   private void showVersionInfo() {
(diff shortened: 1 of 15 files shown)


theute:
For /status what would be the external URL like ? 
Right now the URL is: https://cloud.redhat.com/api/policies/v1/status

PS: We'll have to have the change in mind when releasing to prod or we'll be reported as down (so we'll need to update the probe)

pilhuhn:
The external URL of status would go away and only /status (http://localhost:8080/status ) would be exposed.

theute:
> The external URL of status would go away and only /status (http://localhost:8080/status ) would be exposed.

But we need the status to be exposed externally, since it's used by the NewRelic probe, or did I miss something ?

pilhuhn:
Good point about NewRelic being "on the internet". Will change that back

theute (approved):

PR #161, opened 2020-11-02
Bump to Quarkus 1.9.1

pilhuhn:
Unfortunately Quarkus team has changed some property names, we thus need to change some configuration keys when deploying this.

cc @theute @lahavyuv86 @spinder


theute:
Is there a sibling e2e-deploy PR ? I guess we should merge both "at the same time"

pilhuhn:
Not yet. This sounds like a good idea.

josejulio (approved):

PR #165, opened 2020-11-18
Fix issue with CDI scopes resulting in wrong metric result.

pilhuhn:
Improve testing


theute:
@spinder new metric to alert from, but first to add to the Grafana dashboard ;-)
(Should go to 0 after this is merged)

PR #168, opened 2021-01-05
POL-422 Adds flag to prevent ui-backend from connecting to notificati…

josejulio:
…ons-backend

It's enabled by default, needs to be disabled by setting: `notifications-backend.disable` to `true`


theute:
Sounds good to me merged
@spinder we would need a MR for policies update in prod but notice the new parameters that needs to be set

PR #173, opened 2021-01-18
POL-432 Rollback database changes if any error happened.

josejulio:
- Persisted objects will save automatically, to prevent this
   a RuntimeException should happen between the Transaction or
   manually opt-in for a rollback via TransactionManager


pilhuhn (approved):
I think that makes sense.

PR #183, opened 2021-03-23
NOTIF-161 Remove all email and webhook actions and pulls user preferences from notifications

josejulio:
- email is replaced by notification
 - webhook is removed

A sample of what it does can be seen on: http://sqlfiddle.com/#!15/2907b/12


pilhuhn (approved):

josejulio:
@pilhuhn I'm now removing all the references to `policies-notification` and only accepting `notification` in actions (removing `email`)

PR #226, opened 2021-06-15
POL-527 Add dockerfile and build_deploy.sh

lahavyuv86:


theute:
Can we have build_deploy.sh in `.rhcicd` ?
AFAIK the only change you need to do after that is in your build.yml to declare:
`    build_deploy_script_path: "./.rhcicd/build_deploy.sh"`
along with the gh_repo, quay_org...

lahavyuv86:
@theute Fixed after review

PR #240, opened 2021-07-16
Merge main and test applications.properties files

gwenneg:
The files were identical so the merge is actually a deletion of the `test` file.


josejulio (approved):
lgtm

PR #241, opened 2021-07-20
POL-450 Retrieve policies history entries from PostgreSQL

gwenneg:
Related to https://github.com/RedHatInsights/policies-engine/pull/207.

The policies history is retrieved from the database if the `policies-history.enabled` configuration key is `true`. Otherwise, the history is retrieved from policies-engine with a REST call.

As explained in https://github.com/RedHatInsights/policies-engine/pull/207, here's the migration plan:
- Step 1 (day 1): policies-history is enabled in policies-engine and the engine alerts history is stored in the database. This does not disable the alerts history storage in Infinispan. Both will happen for each alert.
- Step 2 (day 15+): policies-history is enabled in policies-ui-backend. The backend no longer calls the engine through a REST API to retrieve the history. It is read directly from the database.
- Step 3: Change the engine code to stop storing policies history into Infinispan.

If something bad happens with policies-history in any of the projects, we can always revert things to the old way by changing the configuration in both projects. Therefore, the risk of that migration is low.


src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryEntry.java
```diff
@@ -0,0 +1,101 @@
...
+@Entity
+@Table(name = "policies_history")
+public class PoliciesHistoryEntry {
+
+    // The ID is generated by policies-engine.
+    @Id
+    private UUID id;
+
+    @Column(name = "tenant_id")
+    private String tenantId;
```
josejulio, line 19: Should we stop calling it `tenant_id` and start moving to `account_id` instead?

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
```diff
@@ -0,0 +1,162 @@
...
+
+import static com.redhat.cloud.policies.app.model.filter.Filter.Operator.LIKE;
+
+@ApplicationScoped
+public class PoliciesHistoryRepository {
+
+    private static final Logger LOGGER = Logger.getLogger(PoliciesHistoryRepository.class);
+
+    @Inject
+    Session session;
```
josejulio, line 26: The rest of the application uses `Panache`. Is this more effective or better in any way? 
I understand this can be more flexible, but do we do something that can't be done in `Panache` ? A `Session` for custom queries can still be used IIUC.

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
```diff
@@ -0,0 +1,162 @@
...
+        switch (sortColumn) {
+            case "id":
+                return Optional.of("hostId");
+            case "name":
+                return Optional.of("hostName");
+            case "ctime":
+                return Optional.of("ctime");
+            default:
+                // Pager may contain a default sort on `mtime` which is not a PoliciesHistoryEntry field.
+                return Optional.empty();
```
josejulio, line 148:
```suggestion
            case "mtime":
                // Pager may contain a default sort on `mtime` which is not a PoliciesHistoryEntry field.
                return Optional.empty();
            default:
                throw new IllegalArgumentException("Unknown sort column: " + sortColumn);
             
```

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
josejulio: Some of the code could probably be simplified with stuff like:
```
  PanacheQuery<Foo> panacheQuery = find(
            filter.getQuery(),
            pager.getSort(),
            filter.getParameters()
    );
```

and 
```
return new Page<>(
            panacheQuery.list(),
            pager,
            panacheQuery.count()
    );
```

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryEntry.java
gwenneg: This project contains mostly `tenantId` variables and only one `accountId` so I would stick with `tenantId` here.

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
gwenneg: I tried with Panache first, spent a lot of time on it for no good reason and then moved on to something easier to implement (at least for me). 😃 

They say `Panache focuses on making your entities trivial and fun to write in Quarkus` but I'm still looking for the fun part after 2 years knowing Panache. I guess it's great for simple cases when you don't need to build a dynamic query.

Anyway, I don't think it's a problem to have a mix between Panache and plain old HQL/SQL in the same app. It all depends on Hibernate ORM in the end.

gwenneg: If you really prefer Panache, I can give it another try.

josejulio: It's fine, maybe we can find the fun in a separate PR.

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryEntry.java
josejulio: Lets stick to `tenantId` for now then.

josejulio (approved):
lgtm

PR #247, opened 2021-07-26
Replace pr_check.sh with a flag in build_deploy.sh

bennetelli:


theute:
I think we should keep the <testsuite> workaround out of `build_deploy.sh` and keep a separate `pr_check.sh` so that we can add more checks in the future if needed.
pr_check.sh would call `build_deploy.sh testrun"

Also, we shouldn't repeat the `docker build` line, that was the point of making it configurable

theute:
Please don't close PRs as general rule but modify, otherwise next PR on the same topic, someone will miss the context of the discussions/comments of the original PR.
Also feel free to comment if it's a good suggestion or a stupid idea :)

bennetelli:
Okay. Makes sense to me. 

Won't do this again starting from now. (I already opened a new PR for it)

PR #248, opened 2021-07-26
Call build_deploy.sh from pr_check.sh

bennetelli:


bennetelli:
@theute do you mean like this? I started with something like this, but I prefered the other one as it was more readable than this one.

.rhcicd/build_deploy.sh
```diff
@@ -5,24 +5,45 @@ set -exv
...
+if [[ "$1" != "testrun" ]];
+then
+    docker --config="$DOCKER_CONF" push "${IMAGE}:${IMAGE_TAG}"
+    docker --config="$DOCKER_CONF" tag "${IMAGE}:${IMAGE_TAG}" "${IMAGE}:qa"
+    docker --config="$DOCKER_CONF" push "${IMAGE}:qa"
+    docker --config="$DOCKER_CONF" tag "${IMAGE}:${IMAGE_TAG}" "${IMAGE}:latest"
+    docker --config="$DOCKER_CONF" push "${IMAGE}:latest"
+fi
+
+if [ "$1" = "testrun" ];
```
bennetelli, line 40: This part could also be at the end of the pr_check.sh script.

theute: Yes you should move it it's more related to the check than build and deploy

theute (changes requested):

theute (approved):

PR #250, opened 2021-07-29
POL-450 Clean old policies history entries

gwenneg:
The default history retention delay is 14 days. It can be changed with the `policies-history.cleaner.delete-after` conf property which accepts a `Duration` value compatible with the ISO-8601 duration format `PnDTnHnMn.nS`.

The scheduled job execution time doesn't really matter so I scheduled it with the `policies-history.cleaner.period` property which also accepts a `Duration` value. The default value is `P1D` which means once a day.

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
+++ b/src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
@@ -1,0 +1,47 @@
+package com.redhat.cloud.policies.app;
+
+import io.quarkus.scheduler.Scheduled;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.hibernate.Session;
+import org.jboss.logging.Logger;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import javax.transaction.Transactional;
+import java.time.Duration;
+import java.time.Instant;
+
+import static java.util.concurrent.TimeUnit.MINUTES;
+
+@Singleton
+public class PoliciesHistoryCleaner {
+
+    public static final String POLICIES_HISTORY_CLEANER_DELETE_AFTER_CONF_KEY = "policies-history.cleaner.delete-after";
+
+    private static final Logger LOGGER = Logger.getLogger(PoliciesHistoryCleaner.class);
+    private static final Duration DEFAULT_DELETE_DELAY = Duration.ofDays(14L);
+
+    @Inject
+    Session session;
+
+    /**
+     * The policies UI shows a retention time for policies history entries.
+     * This scheduled job deletes from the database the history entries which are older than that retention time.
+     */
+    /*
+     * TODO The scheduling is delayed to prevent an unwanted execution during tests. Remove the delay and set the period
+     * to `disabled` after the Quarkus 2 bump. See https://quarkus.io/guides/scheduler-reference for more details.
+     */
+    @Scheduled(identity = "PoliciesHistoryCleaner", delay = 10L, delayUnit = MINUTES, every = "{policies-history.cleaner.period}")
+    @Transactional
+    public void clean() {
+        Duration deleteDelay = ConfigProvider.getConfig().getOptionalValue(POLICIES_HISTORY_CLEANER_DELETE_AFTER_CONF_KEY, Duration.class)
+                .orElse(DEFAULT_DELETE_DELAY);
+        Instant deleteBefore = Instant.now().minus(deleteDelay);
+        LOGGER.debugf("Policies history purge starting. Entries older than %s will be deleted.", deleteBefore.toString());
+        int deleted = session.createQuery("DELETE FROM PoliciesHistoryEntry WHERE ctime < :ctime")
+                .setParameter("ctime", deleteBefore.toEpochMilli())
+                .executeUpdate();
+        LOGGER.debugf("Policies history purge ended. %d entries were deleted from the database.", deleted);
+    }
+}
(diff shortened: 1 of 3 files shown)


gwenneg:
The scheduled job logs entries like this when the logger is configured at DEBUG level:
```
2021-07-29 09:09:53,737 DEBUG [com.red.clo.pol.app.PoliciesHistoryCleaner] (main) Policies history purge starting. Entries older than 2021-07-15T07:09:53.737743Z will be deleted.
2021-07-29 09:09:53,741 DEBUG [com.red.clo.pol.app.PoliciesHistoryCleaner] (main) Policies history purge ended. 1 entries were deleted from the database.
```

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
```diff
@@ -0,0 +1,47 @@
...
+    Session session;
+
+    /**
+     * The policies UI shows a retention time for policies history entries.
+     * This scheduled job deletes from the database the history entries which are older than that retention time.
+     */
+    /*
+     * TODO The scheduling is delayed to prevent an unwanted execution during tests. Remove the delay and set the period
+     * to `disabled` after the Quarkus 2 bump. See https://quarkus.io/guides/scheduler-reference for more details.
+     */
```
josejulio, line 34: This is not a 2.0 feature, at least since 1.7: https://quarkus.io/version/1.7/guides/scheduler-reference#quarkus-scheduler_quarkus.scheduler.enabled

You could use:
`%test.quarkus.scheduler.enabled=false`

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
```diff
@@ -0,0 +1,47 @@
...
+    public void clean() {
+        Duration deleteDelay = ConfigProvider.getConfig().getOptionalValue(POLICIES_HISTORY_CLEANER_DELETE_AFTER_CONF_KEY, Duration.class)
+                .orElse(DEFAULT_DELETE_DELAY);
+        Instant deleteBefore = Instant.now().minus(deleteDelay);
+        LOGGER.debugf("Policies history purge starting. Entries older than %s will be deleted.", deleteBefore.toString());
+        int deleted = session.createQuery("DELETE FROM PoliciesHistoryEntry WHERE ctime < :ctime")
+                .setParameter("ctime", deleteBefore.toEpochMilli())
+                .executeUpdate();
+        LOGGER.debugf("Policies history purge ended. %d entries were deleted from the database.", deleted);
+    }
```
josejulio, line 46: I suppose we will have the same behavior than with notifications, multiple pods executing the same code. 
In this case is not a big issue as we are not doing anything else than deleting the history.

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
```diff
@@ -0,0 +1,47 @@
...
+    @Transactional
+    public void clean() {
+        Duration deleteDelay = ConfigProvider.getConfig().getOptionalValue(POLICIES_HISTORY_CLEANER_DELETE_AFTER_CONF_KEY, Duration.class)
+                .orElse(DEFAULT_DELETE_DELAY);
+        Instant deleteBefore = Instant.now().minus(deleteDelay);
+        LOGGER.debugf("Policies history purge starting. Entries older than %s will be deleted.", deleteBefore.toString());
+        int deleted = session.createQuery("DELETE FROM PoliciesHistoryEntry WHERE ctime < :ctime")
+                .setParameter("ctime", deleteBefore.toEpochMilli())
+                .executeUpdate();
+        LOGGER.debugf("Policies history purge ended. %d entries were deleted from the database.", deleted);
```
josejulio, line 45: I think this could be an INFO message, is not something that will happen a lot (every day a two lines message) and could be useful to know that it's running

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
gwenneg: I know the conf property to disable the scheduler globally but 2.0 will introduce the possibility to disable each scheduled job individually. Tests will fail in this project if `%test.quarkus.scheduler.enabled=false`, I tried it already.

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
gwenneg: Only this one or both log entries in this method?

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
gwenneg: Yes, it will be executed by all pods, but it shouldn't cause any issue.

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
gwenneg: These are new in Quarkus 2:

> If you wish to disable a specific scheduled method, you can set its cron expression to "off" or "disabled".

> Intervals can be disabled by setting their value to "off" or "disabled". So for example a Property Expression with the default value "off" can be used to disable the trigger if its Config Property has not been set.

src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
josejulio: both please :)

josejulio (approved):
LGTM

PR #259, opened 2021-08-17
POL-450 Fix offset in policies-history queries

gwenneg:
This fixes an offset issue I noticed during tests on stage today.

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
+++ b/src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
@@ -74,9 +74,9 @@ public class PoliciesHistoryRepository {
 
         if (pager.getLimit() > 0) {
             query.setMaxResults(pager.getLimit());
-            if (pager.getOffset() > 0) {
-                query.setFirstResult(pager.getOffset() * pager.getLimit());
-            }
+        }
+        if (pager.getOffset() > 0) {
+            query.setFirstResult(pager.getOffset());
         }
 
         return query.getResultList();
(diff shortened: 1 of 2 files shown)


src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
```diff
@@ -74,9 +74,9 @@ public long count(String tenantId, UUID policyId, Pager pager) {
 
         if (pager.getLimit() > 0) {
             query.setMaxResults(pager.getLimit());
-            if (pager.getOffset() > 0) {
-                query.setFirstResult(pager.getOffset() * pager.getLimit());
-            }
+        }
+        if (pager.getOffset() > 0) {
+            query.setFirstResult(pager.getOffset());
```
josejulio, line 79: This is a global offset? (i.e. not a page?)

gwenneg: The frontend sends a global offset to the backend, but I think the backend was [originally designed](https://github.com/RedHatInsights/policies-ui-backend/blob/20ac41b5790f0e1ad3b47d8ff48c2b2f0c3d084e/src/main/java/com/redhat/cloud/policies/app/model/pager/Pager.java#L75-L78) to receive a page instead of the global offset.

josejulio: Ah. I remember now, we used a page, but the insights API required us to use an offset instead so we changed it.

josejulio (approved):

PR #260, opened 2021-08-19
POL-559 Bump Quarkus to 2.1.3.Final

gwenneg:
The Quarkus path change is documented [here](https://github.com/quarkusio/quarkus/wiki/Migration-Guide-2.0#legacy-redirect-for-non-application-endpoints-such-as-health). The doc is not accurate though: `/health` and `/metrics` work without declaring an absolute path.

Opentracing changes are documented [here](https://github.com/opentracing/opentracing-java/blob/master/CHANGELOG.md#v0330-2019-05-06).

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/auth/RbacFilter.java
+++ b/src/main/java/com/redhat/cloud/policies/app/auth/RbacFilter.java
@@ -18,6 +18,7 @@ package com.redhat.cloud.policies.app.auth;
 
 import com.redhat.cloud.policies.app.RbacServer;
 import io.opentracing.Scope;
+import io.opentracing.Span;
 import io.opentracing.Tracer;
 import io.quarkus.cache.CacheResult;
 import java.io.IOException;
@@ -66,7 +67,8 @@ public class RbacFilter implements ContainerRequestFilter {
     }
 
     long t1 = System.currentTimeMillis();
-    try (Scope ignored = tracer.buildSpan("getRBac").startActive(true)){
+    Span span = tracer.buildSpan("getRBac").start();
+    try (Scope ignored = tracer.scopeManager().activate(span)){
       result = getRbacInfo(user.getRawRhIdHeader());
     } catch (Throwable e) {
       requestContext.abortWith(Response.status(Response.Status.FORBIDDEN).build());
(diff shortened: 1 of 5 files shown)


bennetelli:
My GitHub cli is broken somehow.. gh pr checkout 260 doesn't work.. will check your PR later

src/test/java/com/redhat/cloud/policies/app/NonApplicationRootPathTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+import io.quarkus.test.junit.QuarkusTest;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static io.restassured.http.ContentType.JSON;
+import static io.restassured.http.ContentType.TEXT;
+
+@QuarkusTest
+public class NonApplicationRootPathTest {
```
gwenneg, line 11: I just noticed that RestAssured follows `301` redirections which is not something we want in this class.

I'm testing a fix in policies-engine. I'll commit here in a few minutes if it works.

gwenneg: Fixed.

gwenneg:
PR updated with the latest Quarkus release.

src/main/java/com/redhat/cloud/policies/app/auth/RbacFilter.java
```diff
@@ -76,6 +78,9 @@ public void filter(ContainerRequestContext requestContext) throws IOException {
       if (warnSlowRbac.get() && (t2 - t1) > 500) {
         log.warning("Call to RBAC took " + (t2 - t1) + "ms");
       }
+      if (span != null) {
```
josejulio, line 81: `span` can be null?

gwenneg: Probably not but there are null check on spans in the existing code so I chose the safest path 😄

gwenneg: https://github.com/RedHatInsights/policies-ui-backend/blob/133e13a95e70a9f6ce403b7f52ef7382659c4028/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java#L1054-L1056

The Javadoc isn't really clear about whether or not `span` can be `null` when `tracer.buildSpan("getRBac").start()` is called.

gwenneg: I removed the null-check.

josejulio (approved):

PR #272, opened 2021-09-15
Downgrade Postgres dialect to PostgreSQL82Dialect

gwenneg:
`PostgreSQL10Dialect` is suspected to cause a memory leak. Let's see how it goes with this one.


gwenneg:
Related to https://github.com/RedHatInsights/policies-ui-backend/pull/269

PR #275, opened 2021-09-24
POL-573 Replace run-java.sh script with simple starter

bennetelli:


bennetelli:
@gwenneg @pilhuhn here we are using -XX:-ExitOnOutOfMemoryError -XX:+HeapDumpOnOutOfMemoryError. Should we add HepDumpOnOutOfMemoryError in notifications as well?

gwenneg:
This project has heap dump on OOMError enabled because of a memory leak. This is only temporary. The configuration should be the same as other projects once the leak is fixed.

bennetelli:
> This project has heap dump on OOMError enabled because of a memory leak. This is only temporary. The configuration should be the same as other projects once the leak is fixed.

thanks for the clarification.

pilhuhn:
Also if we don't ship `tar` in the image, the dump can't be pulled off the image :-(

deploy/Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM registry.access.redhat.com/ubi7-minimal
```
pilhuhn, line 1: Do we actually need this file anymore? I think it is a leftover from the early days. And the FROM image is outdated anyway

src/main/docker/Dockerfile.jvm
```diff
@@ -26,17 +24,17 @@ RUN microdnf install openssl curl ca-certificates ${JAVA_PACKAGE} \
...
     && chown 1001:root /deployments \
-    && curl https://repo1.maven.org/maven2/io/fabric8/run-java-sh/${RUN_JAVA_VERSION}/run-java-sh-${RUN_JAVA_VERSION}-sh.sh -o /deployments/run-java.sh \
-    && chown 1001 /deployments/run-java.sh \
-    && chmod 540 /deployments/run-java.sh \
     && echo "securerandom.source=file:/dev/urandom" >> /etc/alternatives/jre/lib/security/java.security
 
-# Configure the JAVA_OPTIONS, you can add -XshowSettings:vm to also display the heap size.
 ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager -XX:-ExitOnOutOfMemoryError -XX:+HeapDumpOnOutOfMemoryError"
 
-COPY --from=build --chown=1001 /home/jboss/target/quarkus-app/ /deployments/
```
bennetelli, line 37: @pilhuhn would the tar be in quarkus-app/ ?

pilhuhn (changes requested):
I think we should get rid of the outdated file while we are here.
Otherwise LGTM

src/main/docker/Dockerfile.jvm
pilhuhn: No. we would need to `dnf install` it. But then the places for it to be used are limited afaiu.

deploy/Dockerfile
bennetelli: I actually thought one of you would use it locally

pilhuhn: Yes, in the past. But even locally let's use the others.

bennetelli: i deleted it. But we still need the clowdapp.yaml file in deploy/?

pilhuhn: You have to ask @lahavyuv86 He seems to have made changes to it recently

pilhuhn (approved):

PR #276, opened 2021-09-28
Stream the check results instead of waiting for completion.

pilhuhn:
This is a draft I had on my disk to stream the results from the verify operation, as it can take long and eventually time out.
By streaming we would at least get some results.


gwenneg:
This is no longer relevant as the `verify` operation no longer exists.

PR #278, opened 2021-10-04
Replace schedule delay workaround with disabled entry when tests running

bennetelli:


src/main/java/com/redhat/cloud/policies/app/PoliciesHistoryCleaner.java
```diff
@@ -28,11 +26,7 @@
      * The policies UI shows a retention time for policies history entries.
      * This scheduled job deletes from the database the history entries which are older than that retention time.
      */
-    /*
-     * TODO The scheduling is delayed to prevent an unwanted execution during tests. Remove the delay and set the period
-     * to `disabled` after the Quarkus 2 bump. See https://quarkus.io/guides/scheduler-reference for more details.
-     */
-    @Scheduled(identity = "PoliciesHistoryCleaner", delay = 10L, delayUnit = MINUTES, every = "{policies-history.cleaner.period}")
+    @Scheduled(identity = "PoliciesHistoryCleaner", every = "{policies-history.cleaner.period}")
```
gwenneg, line 29:
```suggestion
    @Scheduled(identity = "PoliciesHistoryCleaner", every = "${policies-history.cleaner.period}")
```

src/test/resources/application.properties
```diff
@@ -0,0 +1 @@
+policies-history.cleaner.period=disabled
```
gwenneg, line 1: This should be in the `src/main/application.properties` file with a `%test.` prefix:
```suggestion
%test.policies-history.cleaner.period=disabled
```

gwenneg (changes requested):

src/test/resources/application.properties
bennetelli: Ah, that's smarter. Thanks :)

gwenneg: Yeah both approaches work but having a single file is easier to maintain.

gwenneg (approved):
LGTM

PR #282, opened 2021-10-04
PLEASE CHECK THE AUTO FORMATTER RESULTS

bennetelli:


bennetelli:
This is just a test PR to show how auto formatting could work in our project.

With this PR the formatter plugin would format the code when we are building the project. 
PS: It's possible to change the formatter settings.

Would be great to know what you think about it. :)

bennetelli:
In the "Files changed" Tab you can see the results of the formatter plugin.

pilhuhn:
I like the idea of having that formatter. Are the setting compliant with what the IJ rules does (as one of the last PRs introduced that format). 
Ideally running the formatter on the code base after that commit should not change the code format at all.

And: is there a way to create editorconfig files from that format? While we (probably) all use IJ, there are other folks that use Eclipse or VS code

PR #302, opened 2021-10-19
Force javac compiler usage

bennetelli:


bennetelli:
would be great if one of you could checkout master, remove the smallrye-metrics or the jetbrains dependency and do a mvn clean install/verify and check if the jdk internal assertion error appears on your machine as well.

If yes, this issue here fixes this very strange issue.

gwenneg:
This will impact the runtime artifact because of a test issue. I can't approve for now as I don't know exactly all the consequences of adding `forceJavacCompilerUse`. I'll have a look later.

bennetelli:
And it's not only a test classpath issue. I also have this when removing the smallrye-metrics dependency which is a "normal" classpath one.

PR #316, opened 2021-10-27
POL-582 Unhandled error in ui

bennetelli:


src/main/java/com/redhat/cloud/policies/app/model/validation/ActionValidator.java
```diff
@@ -33,7 +33,7 @@ public boolean isValid(String input, ConstraintValidatorContext context) {
             return true;
         }
         for (String action : input.split(";")) {
-            String a = action.toLowerCase().strip();
```
josejulio, line 36: I'm wondering how we got into that bug if this does a toLowerCase here... 
Do we have other means of updating that field?

josejulio: Ah, this is only validating. The stored data is something else.

josejulio (approved):
lgtm

PR #317, opened 2021-10-27
Prepare clowdapp.yaml for the cutover to Clowder deployed stage

gwenneg:
Depends on https://github.com/RedHatInsights/policies-engine/pull/253.

- Dependencies changed to better reflect the real dependencies graph between projects. All dependencies are now mandatory (in opposition to optional before).
- Memory request and limit changed to match the current CI/stage/prod values.
- `policies-ui-backend` no longer owns the `policies_ui_backend` DB from a Clowder perspective. This is necessary to break a Clowder dependencies cycle.
- The `/tmp` volume is now named `rds-client-ca` because it may be used by `clowder-quarkus-config-source` to write the RDS certificate.
- Probes period reduced from `20` to `10` to match the current CI/stage/prod values.
- Several container vars (`APP_NAME`, `ENV_NAME`, `PATH_PREFIX`...) have been removed because they seemed unused. We may need to reintroduce them in case something got broken by that change.

:heavy_check_mark: New ClowdApp template compared with the e2e-deploy template.


theute:
Looks good to me to the best of my limited knowledge

gwenneg:
POL-543

PR #326, opened 2021-11-08
POL-560 Clean code after policies-history migration

gwenneg:


src/test/java/com/redhat/cloud/policies/app/RestApiTest.java
```diff
@@ -655,26 +659,10 @@ void getOnePolicyHistory() {
...
-    @Test
-    void getOnePolicyHistoryBadSort() {
-        String uuid = setupPolicyForHistory();
-
-        try {
-            given()
-                    .header(authHeader)
-                    .contentType(ContentType.JSON)
-                    .when()
-                    .get(API_BASE_V1_0 + "/policies/8671900e-9d31-47bf-9249-8f45698ede72/history/trigger?sortColumn=id")
```
gwenneg, line 667: This `sortColumn` is allowed with Postgres so this test no longer makes sense.

gwenneg, line 667: This `sortColumn` is allowed with Postgres so the test no longer makes sense.

josejulio (approved):
lgtm

PR #332, opened 2021-11-17
POL-601 Limit some features on fedramp

josejulio:
~Missing to test with policies-ui-frontend and see how it behaves~

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
+++ b/src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
@@ -1,0 +1,16 @@
+package com.redhat.cloud.policies.app;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import javax.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class EnvironmentFlags {
+
+    @ConfigProperty(name = "flag.is-fedramp", defaultValue = "false")
+    Boolean isFedramp;
+
+    public Boolean isFedramp() {
+        return isFedramp;
+    }
+}
--- a/src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
+++ b/src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
@@ -1,11 +1,31 @@
 package com.redhat.cloud.policies.app;
 
+import io.quarkus.runtime.Startup;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
 
-import javax.enterprise.context.ApplicationScoped;
+import javax.annotation.PostConstruct;
+import java.lang.reflect.Field;
 
-@ApplicationScoped
+@Startup
 public class EnvironmentFlags {
+
+    private static final Logger LOG = Logger.getLogger(EnvironmentFlags.class);
+
+    @PostConstruct
+    public void init() {
+        try {
+            LOG.info("Environment flags in use:");
+            for (Field field: EnvironmentFlags.class.getDeclaredFields()) {
+                ConfigProperty configAnnotation = field.getAnnotation(ConfigProperty.class);
+                if (configAnnotation != null) {
+                    LOG.infof("  - %s: %s", configAnnotation.name(), field.get(this));
+                }
+            }
+        } catch (IllegalAccessException iae) {
+            throw new RuntimeException("Error getting flags while starting.", iae);
+        }
+    }
 
     @ConfigProperty(name = "flag.is-fedramp", defaultValue = "false")
     Boolean isFedramp;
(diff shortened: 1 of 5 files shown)


src/test/java/com/redhat/cloud/policies/app/UserConfigServiceTest.java
```diff
@@ -75,6 +81,18 @@ void passNotificationResponse() {
...
+    @Test
+    void testFedramp() {
+        when(environmentFlags.isFedramp()).thenReturn(Boolean.FALSE);
+
+        given()
+                .header(authHeader)
+                .when()
+                .get(PREFERENCE_URL)
+                .then()
+                .statusCode(404);
```
josejulio, line 93: Need to check why this is failing on travis but passing on my machine

josejulio:
Tested this with the UI. Seems to behave as we need.

but travis is still mad at me :(

src/test/java/com/redhat/cloud/policies/app/UserConfigServiceTest.java
```diff
@@ -75,6 +81,18 @@ void passNotificationResponse() {
         }
     }
 
+    @Test
+    void testFedramp() {
+        when(environmentFlags.isFedramp()).thenReturn(Boolean.FALSE);
```
josejulio, line 86: Ah.. this test is actually wrong... `mvnw clean test` don't execute it but `mvnw clean verify` does :facepalm:

src/test/java/com/redhat/cloud/policies/app/UserConfigServiceTest.java
bennetelli: what was the reason for the failing test?

src/test/java/com/redhat/cloud/policies/app/UserConfigServiceTest.java
bennetelli: shouldn't test execute it as well?

src/test/java/com/redhat/cloud/policies/app/UserConfigServiceTest.java
josejulio: The environment flag above was set to false, it should be true.

src/test/java/com/redhat/cloud/policies/app/UserConfigServiceTest.java
josejulio: `mvn test` does not run integration tests. I had to run `mvn integration-test` or `mvn install|verify`

src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
```diff
@@ -0,0 +1,36 @@
...
+                    LOG.infof("  - %s: %s", configAnnotation.name(), field.get(this));
+                }
+            }
+        } catch (IllegalAccessException iae) {
+            throw new RuntimeException("Error getting flags while starting.", iae);
+        }
+    }
+
+    @ConfigProperty(name = "flag.is-fedramp", defaultValue = "false")
+    Boolean isFedramp;
```
gwenneg, line 31: Do we need this new config key or could we use [that](https://github.com/RedHatInsights/notifications-backend/blob/1af499740a5478949819353b025851ab347102bf/.rhcicd/clowdapp-backend.yaml#L76-L77) instead (the value is set by Clowder)?

src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
```diff
@@ -0,0 +1,36 @@
...
+    @PostConstruct
+    public void init() {
+        try {
+            LOG.info("Environment flags in use:");
+            for (Field field: EnvironmentFlags.class.getDeclaredFields()) {
+                ConfigProperty configAnnotation = field.getAnnotation(ConfigProperty.class);
+                if (configAnnotation != null) {
+                    LOG.infof("  - %s: %s", configAnnotation.name(), field.get(this));
+                }
+            }
```
gwenneg, line 24: Do you plan on adding more fields later?

src/test/java/com/redhat/cloud/policies/app/UserConfigServiceTest.java
bennetelli: ah. That's differnt to Gradle. Thanks for the hint :)

gwenneg (commented):
I am not allowed to access POL-601. Does this ticket require a special Jira permission?

src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
josejulio: Ah, we could use that one. Do you know what value would take the environment on fedramp ?

src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
josejulio: No plans on specific fields yet, but probably we could have more - I can remove this one for now if that's better.

bennetelli (approved):

src/main/java/com/redhat/cloud/policies/app/EnvironmentFlags.java
gwenneg: I'm not sure yet since Clowder hasn't been used to deploy fedramp until now.

We can merge this PR as-is and simplify it later when the Clowder migration is over.

PR #339, opened 2021-11-25
POL-570 Load last trigger time from Postgres

gwenneg:
I suspect the memory issue we had for a while was not caused by a memory leak but rather by huge triggers lifecycle data returned by the engine when we need to retrieve the last trigger time.

After this PR, ui-backend will no longer retrieve any "last trigger time" from the engine.

Whether my hypothesis is right or not, this will be an improvement anyway.

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
+++ b/src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
@@ -275,26 +275,11 @@ public class PolicyCrudService {
             Pager pager = PagingUtils.extractPager(uriInfo);
             page = Policy.pagePoliciesForCustomer(entityManager, user.getAccount(), pager);
 
-            // Get lastTriggeredTime and add it to the policies
-            String idsToFetch = page.stream().map(p -> p.id.toString()).collect(Collectors.joining(","));
-            try {
-                List<Trigger> list = engine.findTriggersById(idsToFetch, user.getAccount());
-
-                // Put the lastTriggered time (= creation time ) of the alert in a triggerId-alert map
-                Map<String, Long> idTimeMap = new HashMap<>(page.size());
-                for (Trigger trigger : list) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -275,26 +275,11 @@ public Response getPoliciesForCustomer() {
...
-
-                // Put the lastTriggered time (= creation time ) of the alert in a triggerId-alert map
-                Map<String, Long> idTimeMap = new HashMap<>(page.size());
-                for (Trigger trigger : list) {
-                    long lastTriggered = getLastTriggeredFromTriggerLifecycle(trigger);
-                    idTimeMap.put(trigger.id, lastTriggered);
+            for (Policy policy : page) {
+                Long lastTriggerTime = policiesHistoryRepository.getLastTriggerTime(user.getAccount(), policy.id);
+                if (lastTriggerTime != null) {
+                    policy.setLastTriggered(lastTriggerTime);
```
gwenneg, line 281: I will get rid of that iteration later in favor of a SQL join.

theute:
And then it's also one more step in the direction of stop storing trigger history through ISPN I guess

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
pilhuhn: or just `IN ( policyId1, policyId2, ...)`

pilhuhn (approved):

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
gwenneg: You're right. I wrote `join` without really looking at what the other query does :smile:

PR #348, opened 2021-12-07
Stop dumping the heap on OutOfMemoryError

gwenneg:
We no longer need that.


josejulio (approved):

bennetelli:
08:45:00 ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.

Looks like this isn't our issue.

gwenneg:
The build initially failed because of an ephemeral timeout which was caused by a global voluntary slow-down of the ephemeral deployments (for stability purposes).

gwenneg:
I'm merging anyway, we don't need the PR check for this change. I'll update the ephemeral timeout later.

PR #351, opened 2021-12-13
POL-616 Run the DB cleaning job from a CronJob - Phase 1

gwenneg:
Similar to https://github.com/RedHatInsights/notifications-backend/pull/815 which is already running on prod.

### DIFF
--- a/.rhcicd/clowdapp.yaml
+++ b/.rhcicd/clowdapp.yaml
@@ -79,6 +79,65 @@ objects:
           value: ${RBAC_READ_TIMEOUT}
         - name: FLAG_IS_FEDRAMP
           value: ${FLAG_IS_FEDRAMP}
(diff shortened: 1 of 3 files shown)


gwenneg:
Successfully executed on ephemeral:
```
INFO:  2021-12-13 09:40:02.972381+00 Policies history purge starting. Entries older than 14 days will be deleted.
INFO:  2021-12-13 09:40:02.972381+00 Policies history purge ended. 0 entries were deleted from the database.
```

gwenneg:
FYI `SELECT EXTRACT(EPOCH FROM NOW() AT TIME ZONE 'UTC' - INTERVAL '14 days') * 1000;` returns `1638179314755` which can be checked with https://www.epochconverter.com/.

josejulio (approved):

PR #355, opened 2021-12-16
Update Grafana dashboard after clowder migration

bennetelli:


theute:
image

bennetelli:
> 

good catch. thx

PR #358, opened 2021-12-18
Use ENV_NAME to detect the environment

josejulio:


src/main/resources/application.properties
```diff
@@ -84,3 +84,6 @@ quarkus.log.cloudwatch.log-stream-name=policies-ui-backend
 quarkus.log.cloudwatch.level=INFO
 quarkus.log.cloudwatch.access-key-id=placeholder
 quarkus.log.cloudwatch.access-key-secret=placeholder
+
+# Environment
+env.name=""
```
bennetelli, line 89: not sure if it's convention.. should we add a "quarkus." as prefix?

josejulio: I'm trying to load the already existent `ENV_NAME`:
https://quarkus.io/guides/config-reference#environment-variables

gwenneg: No, `quarkus.` is supposed to be a reserved prefix used only for things coming from Quarkus.

bennetelli: ok. was confused as we have that prefix for the cloudwatch extension as wll.

bennetelli (approved):

src/main/java/com/redhat/cloud/policies/app/EnvironmentInfo.java
```diff
@@ -0,0 +1,26 @@
+package com.redhat.cloud.policies.app;
+
+import io.quarkus.runtime.Startup;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.logging.Logger;
+
+import javax.annotation.PostConstruct;
+import java.lang.reflect.Field;
```
gwenneg, line 8: This isn't used.

src/main/java/com/redhat/cloud/policies/app/EnvironmentInfo.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @ConfigProperty(name = "env.name", defaultValue = "")
+    String environmentName;
+
+    @PostConstruct
+    public void init() {
+        LOG.infof("Environment: %s", environmentName);
+    }
+
+    public Boolean isFedramp() {
```
gwenneg, line 23: Nitpick: the unboxing could be avoided:
```suggestion
    public boolean isFedramp() {
```

src/main/resources/application.properties
gwenneg, line 89: There's already a default value in the code, this could be removed.

josejulio: I was getting errors without it. I'll retry.

src/main/java/com/redhat/cloud/policies/app/EnvironmentInfo.java
josejulio: Woaah... why isn't the style check complaining about it?

src/main/resources/application.properties
josejulio: Tests fail to start because of this error: 
```
io.quarkus.runtime.configuration.ConfigurationException: Failed to load config value of type class java.lang.String for: env.name
```

Looks like the empty string counts as "nothing is here, fail". 

So according to this discussion: https://github.com/quarkusio/quarkus/issues/2765
Empty value means not set, adding it as Optional allows the value to be empty without failing. 

I'll update the code.

gwenneg (approved):
LGTM

PR #362, opened 2021-12-22
POL-595 Use quarkus micrometer metrics

bennetelli:


src/main/java/com/redhat/cloud/policies/app/health/ProcSelfStatusExporter.java
```diff
@@ -47,15 +47,35 @@
...
-    long vmStk;
-    long vmLib;
-    long vmData;
-    long vmSize;
-    int threads;
+    @Inject
+    MeterRegistry registry;
+
+    // @Gauge(name = "status.vmHwm", absolute = true, unit = MetricUnits.KILOBYTES, tags = "type=proc")
+    Long vmHwm = registry.gauge("status.vmHwm", 0L);
```
josejulio, line 54: This does't need to be on a `@PostConstruct` method?

bennetelli: Just tested it. Looks like it's not necessary to have it in a postconstruct method

src/main/java/com/redhat/cloud/policies/app/health/ScheduledStatusProducer.java
```diff
@@ -44,11 +47,14 @@
...
 
-    //  // Quarkus only activates this after the first REST-call to any method in this class
-    @Gauge(name = "status_isDegraded", unit = MetricUnits.NONE, absolute = true,
-            description = "Returns 0 if good, value > 0 for number of entries in the status message")
-    int isDegraded() {
-        return StuffHolder.getInstance().getStatusInfo().size();
+    @Inject
+    MeterRegistry registry;
+
+    Gauge degraded;
```
gwenneg, line 53: Do we need a class field for that?

Wouldn't this work?
```
    @PostConstruct
    void init() {
        Gauge.builder("status_isDegraded", () -> StuffHolder.getInstance().getStatusInfo().size()).register(registry);
    }
```

src/main/java/com/redhat/cloud/policies/app/health/ProcSelfStatusExporter.java
```diff
@@ -47,15 +48,31 @@
...
+
+    Long vmHwm;
+    Long vmRss;
+    Long rssAnon;
+    Long rssFile;
+    Long vmStk;
+    Long vmLib;
+    Long vmData;
+    Long vmSize;
+    Integer threads;
```
gwenneg, line 62: I remember seeing `Atomic*` variable in the Micrometer doc. Do we need that here?

src/main/java/com/redhat/cloud/policies/app/health/ProcSelfStatusExporter.java
```diff
@@ -111,51 +128,6 @@ void gather() {
         }
     }
 
-    @Gauge(name = "status.vmHwm", absolute = true, unit = MetricUnits.KILOBYTES, tags = "type=proc")
```
gwenneg, line 114: Can we preserve the tags with the new gauges?

src/test/java/com/redhat/cloud/policies/app/RestApiTest.java
```diff
@@ -53,6 +53,7 @@
 import javax.validation.constraints.NotNull;
 
 import io.restassured.response.Response;
+import io.restassured.response.ValidatableResponse;
```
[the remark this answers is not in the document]
bennetelli: thx

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -92,7 +90,6 @@
 @Path("/api/policies/v1.0/policies")
 @Produces("application/json")
 @Consumes("application/json")
-@SimplyTimed(absolute = true, name = "PolicySvc")
```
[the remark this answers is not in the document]
bennetelli: No, not yet. There is a @Timed annotation in micrometer, but right now we are not using the SimplyTimed data in frafana, that's why I removed it for now. 

My suggestion would be, adding Timed later, if we are missing information/want to add additional graphs.

src/main/java/com/redhat/cloud/policies/app/health/ProcSelfStatusExporter.java
bennetelli: It actually was Atomic* before. Not sure if we need it here. Could be.. but I would need to test it with more than one pod to see if it's necessary or not.

src/main/java/com/redhat/cloud/policies/app/health/ProcSelfStatusExporter.java
bennetelli: We could. This would change it a bit. registry.gauge(..) would become: Gauge.builder("status_isDegraded", () -> StuffHolder.getInstance().getStatusInfo().size())
.tags(...)
.register(registry)

Right now we don't make use of the tags in our grafana graphs. That's why I did it in the simplest possible way (registry.gauge(...) and removed the tags.

src/main/java/com/redhat/cloud/policies/app/health/ScheduledStatusProducer.java
bennetelli: it would. changed.

src/test/java/com/redhat/cloud/policies/app/StatusEndpointTest.java
```diff
@@ -24,6 +24,7 @@
 import static io.restassured.RestAssured.when;
 import static io.restassured.RestAssured.with;
 import static org.hamcrest.core.Is.is;
+import static org.junit.jupiter.api.Assertions.*;
```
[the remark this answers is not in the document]
bennetelli: there is an interesting discussion on stackoverflow regarding star imports pros and cons: https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad :)

src/main/java/com/redhat/cloud/policies/app/health/ProcSelfStatusExporter.java
gwenneg: Grafana is only a subset of the real metrics usage. For instance, I use plenty of metrics from Prometheus that are not in Grafana when I investigate/monitor performances or a bug on stage/prod.

I didn't add nor use (yet) these tags or the `@SimplyTimed` discussed above so I don't have a strong feeling about keeping or removing them, but they were added for a reason. I'm not sure this migration should alter any existing metric (as long as it can be migrated).

bennetelli: I asked Heiko. He added it.

We can either replace it with @Timed annotation or use Antonin Stefanutti's extension, put the @Metrics annotation on the class and the Timed annotation on the respective method.

PR #401, opened 2022-04-07
Log exceptions cause and stack trace

gwenneg:
Prod is unstable and the current code is hiding the real issue behind a `Connection to backend-engine failed. Please retry later` message.


theute:
LGTM

PR #415, opened 2022-04-25
POL-595 Use micrometer metrics

bennetelli:
How that old version worked is still very confusing to me.


src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -533,7 +534,7 @@ public Response deletePolicies(List<UUID> uuids) {
         }
 
         // TODO Temp counter used to investigate the engine instability, remove ASAP.
```
bennetelli, line 536: @gwenneg can we remove the counters already?

PR #420, opened 2022-05-06
POL-649 Introduce lightweight engine integration

gwenneg:
Related to https://github.com/RedHatInsights/policies-engine/pull/375.

This PR introduces a new configuration key that can be used to replace all calls to the old `policies-engine` REST API from `TriggersHandler` with new a new REST API dedicated to the lightweight engine:

```
lightweight-engine.enabled=true
```
ℹ️ If that config key is set to `false` (default value), the old behavior will remain unchanged. This PR is therefore safe to merge as it will not affect prod until the new mode is enabled from configuration.

Two operations from `AdminService` are unavailable (503 status) when the lightweight engine is enabled:
- `GET /admin/verify` (finds the orphan triggers)
- `POST /admin/sync` (syncs the triggers between the ui-backend DB and the engine PV)

▶️ Before this PR:
- `ui-backend` and `engine` send/receive full triggers that are used to validate/retrieve/store/update triggers and their conditions.
- `ui-backend` sends a trigger ID to `engine` to delete a trigger.

▶️ After this PR:
- Conditions are validated with a new dedicated `engine` REST endpoint which accepts the condition as the request body (text).
- All other calls from `ui-backend` to `engine` are done with a single REST endpoint which accepts a collection of triggers IDs as the request body (JSON). Each time that endpoint is called, the `engine` will try to reload all triggers identified by the triggers IDs passed in the request. This is explained in details in the `LightweightEngine` javadoc in this PR.

⚠️ Check the `Hide whitespace` option for an easier review.


gwenneg:
ℹ️ I'm starting local tests with both `engine` and `ui-backend` PRs. They can be reviewed in the meantime.

src/main/java/com/redhat/cloud/policies/app/LightweightEngine.java
```diff
@@ -0,0 +1,43 @@
...
+
+// TODO POL-649 Retries on all calls: https://quarkus.io/guides/smallrye-fault-tolerance
+// TODO POL-649 Exception mapper?
+@Path("/lightweight-engine")
+@RegisterRestClient(configKey = "engine")
+public interface LightweightEngine {
+
+    // TODO POL-649 The validation errors need to be properly forwarded to the browser.
+    /**
+     * Validates a condition with Hawkular.
```
bennetelli, line 24: If you want, you could add an example of an condition. Maybe it's clear for everyone using this endpoint and hawkular besides me :D but I would not know how such a condition would look like at the moment

src/main/java/com/redhat/cloud/policies/app/rest/AdminService.java
```diff
@@ -61,6 +62,9 @@ public class AdminService {
 
     private final Logger log = Logger.getLogger(this.getClass().getSimpleName());
 
+    @Inject
+    LightweightEngineConfig lightweightEngineConfig;
```
bennetelli, line 66: ok. maybe that is one of the reasons why it is not in the engine but in a separate config class instead.

src/main/java/com/redhat/cloud/policies/app/rest/AdminService.java
```diff
@@ -135,6 +139,10 @@ public Response setAdminDown(@QueryParam("status") Optional<String> status) {
     @Transactional
     public Response syncToEngine(@QueryParam("token") String token) {
 
+        if (lightweightEngineConfig.isEnabled()) {
+            return Response.status(503).entity("Sync unavailable when the lightweight engine is enabled").build();
```
bennetelli, line 143: a very simple unit test would be great so we have something like an insurance when we change something somewhere in the code.

src/main/java/com/redhat/cloud/policies/app/rest/AdminService.java
```diff
@@ -169,6 +177,10 @@ public Response syncToEngine(@QueryParam("token") String token) {
     @Path("/verify")
     public Response findOrphans() {
 
+        if (lightweightEngineConfig.isEnabled()) {
+            return Response.status(503).entity("Verify unavailable when the lightweight engine is enabled").build();
```
bennetelli, line 181: a very simple unit test would be great so we have something like an insurance when we change something somewhere in the code.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -512,27 +531,41 @@ public Response deletePolicies(List<UUID> uuids) {
         // TODO Temp counter used to investigate the engine instability, remove ASAP.
         registry.counter("policies_ui_deletePolicies", new Tag("account", user.getAccount())).inc();
 
+        // TODO POL-649 Do we really need to return this?
```
josejulio, line 534: Yes, as the operation was not a single transaction, some policies could fail to be deleted. The UI needs to know which was deleted.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -512,27 +531,41 @@ public Response deletePolicies(List<UUID> uuids) {
...
-                } catch (NotFoundException nfe) {
-                    // Engine does not have it - we can delete anyway
-                    deletedOnEngine = true;
-                } catch (Exception e) {
-                    log.warn("Deletion on engine failed", e);
-                }
-                if (deletedOnEngine) {
+                if (policy != null) {
                     policy.delete();
+                    reallyDeleted.add(uuid);
```
josejulio, line 544: are you sure it was really really really deleted? :grin:

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -994,4 +1063,25 @@ private Response isNameUnique(Policy policy) {
...
     }
+
+    private Response reloadEngineTriggers(String accountId, Set<UUID> policyIds, Supplier<ResponseBuilder> successResponseSupplier) {
+        try {
+            lightweightEngine.reloadTriggers(accountId, policyIds);
+        } catch (Throwable t) {
+            log.error("Failed to reload engine triggers", t);
+            try {
+                // Any change previously done to the DB data during the current transaction will be rolled back.
+                transactionManager.setRollbackOnly();
```
josejulio, line 1074: How does this work? 
The engine takes the ids, but are the values available to be read by the engine if they are not commited? 
Also - what happens if the engine is down - do we really need it for the policy CRUD ?

src/main/java/com/redhat/cloud/policies/app/rest/AdminService.java
gwenneg: One of the reasons explained above, yes 😃

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -406,30 +415,35 @@ public Response storePolicy(@QueryParam("alsoStore") boolean alsoStore, @NotNull
...
 
-        // Basic validation was successful, so try to persist.
-        // This may still fail du to unique name violation, so
-        // we need to check for that.
-        UUID id;
-        try {
-            FullTrigger trigger = new FullTrigger(policy);
+        if (lightweightEngineConfig.isEnabled()) {
+            policy.store(user.getAccount(), policy);
+            // TODO POL-649 The old code returned a policy location. Was it used by the frontend?
```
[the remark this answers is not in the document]
josejulio: I think it is: https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/Response.html#created-java.net.URI-
We don't use it.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
gwenneg: I learned from the Camel integration in notifications 😄

src/main/java/com/redhat/cloud/policies/app/LightweightEngine.java
gwenneg: Done.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
gwenneg: TODO removed.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -691,22 +748,30 @@ public Response updatePolicy(@QueryParam("dry") boolean dryRun, @PathParam("poli
...
                     storedPolicy.setMtimeToNow();
 
                     existingTrigger.updateFromPolicy(storedPolicy);
-                    try {
-                        engine.updateTrigger(storedPolicy.id, existingTrigger, false, user.getAccount());
-                    } catch (Exception e) {
-                        transactionManager.setRollbackOnly();
-                        return Response.status(400, e.getMessage()).entity(getEngineExceptionMsg(e)).build();
+                    if (lightweightEngineConfig.isEnabled()) {
+                        return reloadEngineTrigger(user.getAccount(), storedPolicy.id, () -> Response.ok(storedPolicy));
```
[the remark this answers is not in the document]
gwenneg: I haven't checked yet, but it should. We are within the boundaries of a JTA transaction and modifying an entity managed by Hibernate. The changes are supposed to be comitted if the transaction ends without error. It was already working like that before my PR.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
gwenneg: Good question! I haven't started the local tests yet but thanks to your comment I can already tell that there will be an issue here 😄 Not a big one though. I'll probably move the transactions boundaries as everything needs to be comitted in `ui-backend` before we call `engine`.

src/main/java/com/redhat/cloud/policies/app/rest/AdminService.java
gwenneg: This is a perfect occasion to show one of the benefits of `LightweightEngineConfig` 😃

gwenneg: Test added.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
gwenneg: This may actually encourage me to do something I had in mind and didn't implement yet. I may have been wrong writing on Slack that there wouldn't be any major code change. 😄

gwenneg:
ℹ️ I changed the triggers sync logic between engine and ui-backend.

See https://github.com/RedHatInsights/policies-engine/pull/375#issuecomment-1124665417 for more details.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -607,33 +651,47 @@ public Response setEnabledStateForPolicies(@QueryParam("enabled") boolean should
...
-                            engine.enableTrigger(storedPolicy.id, user.getAccount());
-                        } else {
-                            engine.disableTrigger(storedPolicy.id, user.getAccount());
+                    storedPolicy.isEnabled = shouldBeEnabled;
+                    storedPolicy.setMtimeToNow();
+                    storedPolicy.persist();
+                    changed.add(uuid);
+                }
+            }
+            accountLatestUpdateRepository.setLatestToNow(user.getAccount());
```
josejulio, line 664: I guess this could count as an optimization, not something i would expect to happen often so feel free to ignore it

```suggestion
                    if (storedPolicy.isEnabled != shouldBeEnabled) {
                        storedPolicy.isEnabled = shouldBeEnabled;
                        storedPolicy.setMtimeToNow();
                        storedPolicy.persist();
                        changed.add(uuid);
                    }
                }
            }
            if (!changed.isEmpty()) {
                accountLatestUpdateRepository.setLatestToNow(user.getAccount());
            }
```

src/main/java/com/redhat/cloud/policies/app/lightweight/LightweightEngineConfig.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public boolean isEnabled() {
+        return enabled;
+    }
+
+    /**
+      * <b>/!\ WARNING /!\</b> Do not use this method from runtime code.
+      */
+    public void overrideForTest(boolean enabled) {
```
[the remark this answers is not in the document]
gwenneg: Keep in mind that this entire class will disappear soon. I don't think we need reflection here 😃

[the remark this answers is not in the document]
bennetelli: offtopic question, but is Quarkus able to work with reflection things out of the box? As far as I remember you had to set a property or something like that to be able to work with it and create a native executable even when there is reflaction included? 

I am just interested :D

gwenneg:
Self-review complete.

gwenneg:
Comment have been answered, issues have been fixed and @josejulio approved the PR in a discussion on Slack. I'm merging and deploying on stage.

PR #421, opened 2022-05-09
Remove SettingsValues class

bennetelli:


pilhuhn:
I think that is ok. Wasn't it used in the past? I guess that is now com.redhat.cloud.policies.app.NotificationSystem.UserPreferences

pilhuhn (approved):

bennetelli:
> I think that is ok. Wasn't it used in the past? I guess that is now com.redhat.cloud.policies.app.NotificationSystem.UserPreferences

Yes, that's what Thomas was thinking of as well.

PR #430, opened 2022-05-18
POL-649 Remove old engine integration

gwenneg:
The old engine no longer exists. This PR removes all the integration code that was targeting it.


src/test/java/com/redhat/cloud/policies/app/PolicyCrudServiceTest.java
```diff
@@ -65,44 +60,16 @@ void test() {
         assertEquals(5, history.getJsonObject("meta").getInteger("count"));
     }
 
-    private UUID createPolicy() {
+    @Transactional
```
bennetelli, line 63: why is this needed now?

gwenneg: Because of the SQL query that is executed in the new method body. The old version didn't execute a SQL query.

josejulio (approved):
lgtm

PR #434, opened 2022-05-31
POL-650 Add org id migration job

bennetelli:


.rhcicd/clowdapp.yaml
```diff
@@ -17,6 +17,37 @@ objects:
...
+            image: quay.io/cloudservices/tenant-utils:latest
+            command:
+              - ./org-id-column-populator
+              - -C
+              - -a
+              - account_id
+              - -o
+              - org_id
+              - -t
+              - account_latest_update
```
bennetelli, line 32: that splitting table (account_latest_update and org_id_latest_update) looks like a problem here, @gwenneg 

Is it an option to add the orgId to the account_latest_update even if it's crap from a semantical perspective? If not I would need to ask Derek if he could tweek his tool a bit.

josejulio: I guess we don't need to migrate anything here. 

It's more of a runtime thing so we are fine without migrating this data.

josejulio: @gwenneg can confirm, but IIRC this just tell an already running engine if it needs to reload the data from the database. A new engine will load everything regardless of this table.
The only requirement for it to work is that the policies-engine and the policies-ui-backend are using the same config (both org id or both account id)

josejulio: In the event that we needed to migrate this table (I doubt) we could just add the orgId to the existing table and rename afterwards.

bennetelli: as long as we set the flags correctly in stage and prod, this should be no problem.

bennetelli: > @gwenneg can confirm, but IIRC this just tell an already running engine if it needs to reload the data from the database. A new engine will load everything regardless of this table. The only requirement for it to work is that the policies-engine and the policies-ui-backend are using the same config (both org id or both account id)



as long as we set the flags correctly in stage and prod, this should be no problem.

bennetelli:
Who closed this PR?

gwenneg:
I'll review this PR today.

gwenneg:
Changes moved to #447.

PR #445, opened 2022-06-13
POL-650 Add orgId

bennetelli:


pom.xml
```diff
@@ -114,7 +114,7 @@
     <dependency>
       <groupId>io.quarkiverse.logging.cloudwatch</groupId>
       <artifactId>quarkus-logging-cloudwatch</artifactId>
-      <version>4.3.0</version>
+      <version>4.3.1</version>
```
gwenneg, line 117: Could you please revert that change?

src/main/java/com/redhat/cloud/policies/app/auth/RhIdPrincipal.java
```diff
@@ -25,20 +25,18 @@ public class RhIdPrincipal implements Principal {
 
     private String name;
     private String account;
+    private String orgId;
     private boolean canReadPolicies;
     private boolean canWritePolicies;
     private String rawRhId;
 
-    public RhIdPrincipal() {
-    }
```
gwenneg, line 33: This could be needed by some code building an instance from reflection code. Or not...

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
```diff
@@ -84,12 +84,14 @@ public List<PoliciesHistoryEntry> find(String tenantId, UUID policyId, Pager pag
 
     private static String addFiltersConditions(String hql, List<Filter.FilterItem> filterItems) {
         // The filters from the pager are added to the HQL query.
+        StringBuilder hqlBuilder = new StringBuilder(hql);
         for (Filter.FilterItem filterItem : filterItems) {
             String entityFieldName = getEntityFieldName(filterItem);
             String operator = getHqlOperator(filterItem);
             // To be consistent with the previous implementation, the condition is always case-insensitive.
-            hql += " AND LOWER(" + entityFieldName + ")" + operator + ":" + entityFieldName;
+            hqlBuilder.append(" AND LOWER(").append(entityFieldName).append(")").append(operator).append(":").append(entityFieldName);
```
gwenneg, line 92: `StringBuilder` is slower in modern Java than the `+` operator when the number of iterations is reasonable (it is here). The opposite was true in the past.

src/main/resources/db/migration/V19__POL-650_add_org_id_columns.sql
```diff
@@ -1,15 +0,0 @@
-CREATE TABLE org_id_latest_update (
```
gwenneg, line 1: Why is this here? This PR shouldn't contain any SQL file changes.

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
```diff
@@ -84,12 +84,14 @@ public List<PoliciesHistoryEntry> find(String tenantId, UUID policyId, Pager pag
```
gwenneg, line 84: There are several method based on `tenantId` in this class that need to be duplicated to use the `orgId` instead. Calling one version or the other should be determined by the boolean from the configuration.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -249,7 +249,12 @@ public Response getPoliciesForCustomer() {
         Page<Policy> page;
         try {
             Pager pager = PagingUtils.extractPager(uriInfo);
-            page = Policy.pagePoliciesForCustomer(entityManager, user.getAccount(), pager);
+
+            if (user.getOrgId() != null) {
+                page = Policy.pagePoliciesForCustomer(entityManager, user.getOrgId(), pager);
+            } else {
+                page = Policy.pagePoliciesForCustomer(entityManager, user.getAccount(), pager);
```
gwenneg, line 256: Most methods from `Policy` need to be duplicated. The old ones should continue dealing with the account while the new ones should deal with the `orgId`. The `account` and the `orgId` are different things (and will coexist for a while). They should never be mixed at any place in the code.

gwenneg: The `Policy` class also needs the `orgId` field, it is currently missing.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -381,14 +396,20 @@ public Response storePolicy(@QueryParam("alsoStore") boolean alsoStore, @NotNull
...
-        policy.store(user.getAccount(), policy);
-        accountLatestUpdateRepository.setLatestToNow(user.getAccount());
+        if (user.getOrgId() != null) {
+            policy.store(user.getOrgId(), policy);
+            accountLatestUpdateRepository.setLatestOrgIdToNow(user.getOrgId());
+        }
+        if (user.getAccount() != null) {
+            policy.store(user.getAccount(), policy);
+            accountLatestUpdateRepository.setLatestToNow(user.getAccount());
+        }
```
gwenneg, line 406: There an issue here as it will store the same policy 2 times.

gwenneg (changes requested):

src/main/resources/db/migration/V19__POL-650_add_org_id_columns.sql
bennetelli: It deletes the script. You wanted this to be deleted.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
bennetelli: fuck yes. We should add a technical debt task. There are so many tests missing everywhere.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
bennetelli: done

src/main/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepository.java
bennetelli: done

src/main/java/com/redhat/cloud/policies/app/model/Policy.java
```diff
@@ -270,20 +342,6 @@ public static FilterableColumn fromName(String columnName) {
...
-        for (int i = 0; i < columns.size(); i++) {
-            Sort.Column column = columns.get(i);
-            if (i > 0)
-                sb.append(" , ");
-            sb.append(column.getName());
-            if (column.getDirection() != Sort.Direction.Ascending)
-                sb.append(" DESC");
-        }
-        return sb.toString();
-    }
```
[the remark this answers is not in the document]
bennetelli: Yes it was. I double checked it. Unused in intelliJ and unused in text search

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -817,24 +906,46 @@ public Response getTriggerHistoryForPolicy(@PathParam("id") UUID policyId) {
...
     }
 
     private Response isNameUnique(Policy policy) {
-        Policy tmp = Policy.findByName(user.getAccount(), policy.name);
+        Policy tmp;
+        if (user.getOrgId() != null) {
+            tmp = Policy.findByNameOrgId(user.getOrgId(), policy.name);
+        } else {
+            tmp = Policy.findByName(user.getAccount(), policy.name);
+        }
```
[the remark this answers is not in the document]
bennetelli: this should do the job now

gwenneg:
The jboss logger was removed from all projects recently.

We're using the Quarkus `Log` API now (which has the exact same signatures than the jboss one).

bennetelli:
> The jboss logger was removed from all projects recently.
> 
> We're using the Quarkus `Log` API now (which has the exact same signatures than the jboss one).

and why is it not working in GitHub Actions anymore?

src/main/resources/db/migration/V19__POL-650_add_org_id_columns.sql
bennetelli: reverted that one above and removed the other one

bennetelli:
> > The jboss logger was removed from all projects recently.
> > We're using the Quarkus `Log` API now (which has the exact same signatures than the jboss one).
> 
> and why is it not working in GitHub Actions anymore?

I still don't understand why it's failing here and working elsewhere

gwenneg:
The failure is caused by something very counter-intuitive from GitHub: this PR is built with your changes applied on an old version of `master`. GitHub always works like that but it's usually harmless and we don't notice it.

Rebasing the PR on the latest `master` should help getting rid of the compilation issue.

bennetelli:
> The failure is caused by something very counter-intuitive from GitHub: this PR is built with your changes applied on an old version of `master`. GitHub always works like that but it's usually harmless and we don't notice it.
> 
> Rebasing the PR on the latest `master` should help getting rid of the compilation issue.

i did not find any noticable changes on latest master. Anyway. We should try to minimize the time a branch "lives" in the future.

gwenneg:
This is the diff from this PR:

image

The `Logger` highlighted in red no longer exists on the latest `master`.

bennetelli:
> This is the diff from this PR:
> 
> image
> 
> The `Logger` highlighted in red no longer exists on the latest `master`.

should be fixed now

gwenneg (approved):
LGTM

PR #454, opened 2022-06-20
Use non deprecated sentry logging

bennetelli:


pom.xml
```diff
@@ -117,6 +113,12 @@
       <version>4.3.1</version>
     </dependency>
 
+    <dependency>
+      <groupId>io.quarkiverse.loggingsentry</groupId>
+      <artifactId>quarkus-logging-sentry</artifactId>
+      <version>1.0.1</version>
```
josejulio, line 119: I'm assuming quarkiverse stuff gets its own version?

bennetelli: yes. the quarkiverse dependencies can have every version they want to. cloudwatch-logging for example is on version 4.3.1.
The release cycle is independent from the quarkus-bom and quarkus in general. Which is good and bad at the same time.

PR #464, opened 2022-07-06
POL-644 Reject bad values for booleans.

pilhuhn:
Only empty, true, false should be accepted.


bennetelli (approved):
tried to test the exception message, but it seems to be non present in the response, so LGTM

PR #468, opened 2022-07-07
Stop changing the customerId/orgId on policy update

gwenneg:


src/main/java/com/redhat/cloud/policies/app/model/Policy.java
```diff
@@ -255,14 +255,11 @@ public void delete(Policy policy) {
     }
 
     public void populateFrom(Policy policy) {
-        this.id = policy.id;
```
gwenneg, line 258: `this` comes from the DB and already has an `id` value. It doesn't make sense to change it here.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -584,7 +584,6 @@ public Response updatePolicy(@QueryParam("dry") boolean dryRun, @PathParam("poli
                 // All is good, we can now do the real work
                 try {
                     storedPolicy.populateFrom(policy);
-                    storedPolicy.customerid = user.getAccount();
```
gwenneg, line 587: `storedPolicy` comes from the DB and already has a `customerid` value. It doesn't make sense to change it here.

josejulio (approved):
lgtm

PR #479, opened 2022-07-21
Bump Quarkus to 2.11 and then log the org id in access log.

pilhuhn:
Also log orgId on slow rbac

This will then look like

`2022-07-20 17:33:26,462 INFO  [access_log] (executor-thread-0) "127.0.0.1 - hrupp 20/Jul/2022:17:33:26 -0400 "GET /api/policies/v1.0/policies HTTP/1.1" 403 - orgId=1234567"
`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 
   <properties>
     <enforcer-plugin.version>3.1.0</enforcer-plugin.version>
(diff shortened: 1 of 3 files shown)


theute:
<3

josejulio (approved):

PR #497, opened 2022-08-18
POL-650 Make Org ID mandatory and account optional in DB

gwenneg:
Checked with `gabi` today:
| Query | Stage | Prod |
| - | - | - |
| select count(*) from policy where org_id is null | 0 | 0 |
| select count(*) from policies_history where org_id is null | 0 | 0 |


src/main/java/com/redhat/cloud/policies/app/auth/IncomingRequestFilter.java
```diff
@@ -136,9 +136,10 @@ private XRhIdentity determineXRhIdentity(String xrhid_header) {
         // Basic sanity check
         XRhIdentity rhIdentity = xrhid.get();
         if (rhIdentity.getUsername() == null || rhIdentity.getUsername().isEmpty() ||
-                rhIdentity.identity.accountNumber == null || rhIdentity.identity.accountNumber.isEmpty()
+                rhIdentity.identity.accountNumber == null || rhIdentity.identity.accountNumber.isEmpty() ||
```
josejulio, line 139: Should we remove this check? IIRC some users won't have an account id

gwenneg: Done

src/main/resources/db/migration/V21__POL-650_org_id_constraints.sql
```diff
@@ -0,0 +1,13 @@
+DROP INDEX pol_cus_nam_idx;
+
+ALTER TABLE policy
+    DROP CONSTRAINT policy_pkey,
+    ALTER COLUMN customerid DROP NOT NULL,
```
josejulio, line 5: isn't the  customer_id the account_id field? Are we removing it?

src/main/resources/db/migration/V21__POL-650_org_id_constraints.sql
```diff
@@ -0,0 +1,13 @@
...
+    ALTER COLUMN org_id SET NOT NULL,
+    ADD CONSTRAINT policy_pkey PRIMARY KEY (org_id, id);
+
+CREATE UNIQUE INDEX policy_org_id_name_idx ON policy (org_id, name);
+
+ALTER TABLE policies_history
+    ALTER COLUMN tenant_id DROP NOT NULL,
```
josejulio, line 12: same as above

src/main/resources/db/migration/V21__POL-650_org_id_constraints.sql
gwenneg: Yes, this is where the account ID is stored, but this script only removes a `NOT NULL` constraint, not the field itself.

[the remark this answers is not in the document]
josejulio: ah - nevermind, its removing the NOT NULL

src/main/resources/db/migration/V21__POL-650_org_id_constraints.sql
josejulio: nevermind this one too :D

josejulio (approved):
lgtm

PR #499, opened 2022-08-23
Bump Maven version and maven-wrapper

gwenneg:
`maven-wrapper` is now available under the `org.apache.maven.wrapper` group in [Maven Central](https://search.maven.org/search?q=g:org.apache.maven.wrapper%20AND%20a:maven-wrapper).

This PR also aligns the Maven version with Quarkus 2.11.2.Final.


pilhuhn (approved):
LGTM (besides the fact that I still don't grok why ppl find the wrapper awesome )

gwenneg:
I wouldn't describe it as awesome, but making sure everyone builds a project with the exact same version can help avoid build issues resulting from different versions of Maven being used.

gwenneg:
The PR check failure is unrelated, IQE tests were already broken on ephemeral.

pilhuhn:
That "same version" is IMO rather/better done by the enforcer plugin.

gwenneg:
Can it enforce a _maximum_ Maven version?

gwenneg:
It looks like it does.

There's another cool thing about the wrapper: you don't have to bother downloading version x.y.z of Maven because your project needs it :smiley:

pilhuhn:
It can enforce a version range. See https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

And then one does not/can't enforce using the wrapper. And the wrapper is mutually exclusive with `mvnd`.

josejulio:
> There's another cool thing about the wrapper: you don't have to bother installing version x.y.z of Maven because your project needs it smiley

I think this is the biggest selling point for me

PR #537, opened 2023-03-07
Bump java version to 17

josejulio:


pom.xml
```diff
@@ -11,7 +11,7 @@
   <properties>
     <enforcer-plugin.version>3.1.0</enforcer-plugin.version>
     <maven-minimum-version>3.8.6</maven-minimum-version>
-    <java-version>11</java-version>
+    <java-version>17</java-version>
```
gwenneg, line 14: There are two other references to Java 11 below (lines 491 and 492). I don't know if they need to be updated.

josejulio: Updated to use `${java.version}` and forcing a run of the sonar-scan.

gwenneg (approved):
LGTM

PR #551, opened 2023-05-05
Adding conditional to build deploy script in security compliance branch

xxlhacker:
Adding conditional logic to the build deploy script for the security compliance branch.


casey-williams-rh (approved):
👍

PR #558, opened 2023-07-17
refactor(repository): use native SQL for history

vkrizan:
Refactors `PoliciesHistoryRepository` to use native SQL to support Postgresql-specific queries (like JSON lookups).

RHINENG-1191

### DIFF
--- a/src/test/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepositoryTest.java
+++ b/src/test/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepositoryTest.java
@@ -6,6 +6,8 @@ import com.redhat.cloud.policies.app.model.pager.Pager;
 import io.quarkus.test.common.QuarkusTestResource;
 import io.quarkus.test.junit.QuarkusTest;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.BeforeAll;
 
 import javax.inject.Inject;
 import java.util.List;
@@ -21,6 +23,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @QuarkusTest
 @QuarkusTestResource(TestLifecycleManager.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class PoliciesHistoryRepositoryTest {
 
     private static final String TENANT_ID_1 = "tenant-id-1";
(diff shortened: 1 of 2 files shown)


gwenneg:
Quarkus 3 comes with Hibernate ORM 6 which supports the JSONB type. That may be something to consider instead of transforming the queries to native SQL.

vkrizan:
@gwenneg does Hibernate 6 support JSONB lookups in HQL? What I've found is that there is no support for it in HQL. It complains about `@>` operator in HQL ver. 5.

Upgrading to Quarkus 3 might be a heavier lift than changing to native SQL, which most of our team "speak".

vkrizan:
@gwenneg Per discussion, this would be a much heavier lift at the moment. Would you have time to review this refactoring. Thank you!

src/test/java/com/redhat/cloud/policies/app/model/history/PoliciesHistoryRepositoryTest.java
```diff
@@ -40,23 +41,25 @@ class PoliciesHistoryRepositoryTest {
...
-         */
-        PoliciesHistoryEntry historyEntry1 = helper.createPoliciesHistoryEntry(TENANT_ID_1, ORG_ID_1, POLICY_ID_1, HOST_ID_1, HOST_NAME_1, 1L);
-        PoliciesHistoryEntry historyEntry2 = helper.createPoliciesHistoryEntry(TENANT_ID_2, ORG_ID_2, POLICY_ID_1, HOST_ID_1, HOST_NAME_1, 2L);
-        PoliciesHistoryEntry historyEntry3 = helper.createPoliciesHistoryEntry(TENANT_ID_2, ORG_ID_2, POLICY_ID_2, HOST_ID_1, HOST_NAME_1, 3L);
-        PoliciesHistoryEntry historyEntry4 = helper.createPoliciesHistoryEntry(TENANT_ID_2, ORG_ID_2, POLICY_ID_2, HOST_ID_1, HOST_NAME_2, 3L);
-        PoliciesHistoryEntry historyEntry5 = helper.createPoliciesHistoryEntry(TENANT_ID_2, ORG_ID_2, POLICY_ID_2, HOST_ID_2 + "-foo", HOST_NAME_2, 4L);
-        PoliciesHistoryEntry historyEntry6 = helper.createPoliciesHistoryEntry(TENANT_ID_2, ORG_ID_2, POLICY_ID_2, HOST_ID_2, "bar-" + HOST_NAME_2, 5L);
+    private static List<PoliciesHistoryEntry> insertedEntries;
+
+    @BeforeEach
```
josejulio, line 46: Could be replaced with a @BeforeAll annotation and we can remove the if below

vkrizan: It needs to be static and so does the `helper` (`PoliciesHistoryTestHelper`). Injection of the helper, however, happens on initialized class. Any ideas how to overcome it?

josejulio: You can add `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` to the class. Meaning that it will run all the tests on a single instance instead of  creating a new one per test. This allows to use the BeforeAll as a instance method.

It seems OK given that you don't have any instance fields or local state.

vkrizan: Done. Thank you.

josejulio (approved):
lgtm

PR #561, opened 2023-07-21
feat(auth): RBAC attribute filters support

vkrizan:
Adds parsing of `attributeFilter` from RBAC access responses.

Refactors tests to use Jackson, the same as the RestClient uses.

_Hint: review by commits_

RHINENG-1304


josejulio (approved):
lgtm

PR #563, opened 2023-08-01
build: switch from json-b to jackson

josejulio:


src/main/java/com/redhat/cloud/policies/app/EngineResponseExceptionMapper.java
```diff
@@ -55,8 +54,9 @@ public RuntimeException toThrowable(Response response) {
     private Msg getBody(Response response) {
         String msg = response.readEntity(String.class);
         if (msg != null && !msg.isEmpty()) {
-            try (Jsonb jsonb = JsonbBuilder.create()) {
-                Map<String, String> errorMap = jsonb.fromJson(msg, HashMap.class);
+            ObjectMapper objectMapper = new ObjectMapper();
```
josejulio, line 57: This `ObjectMapper` could be injected. Unsure if this class accepts `@Inject`s but maybe it will do as I see it has a `@Priority` annotation. didn't test.

src/main/java/com/redhat/cloud/policies/app/model/Policy.java
```diff
@@ -86,19 +89,21 @@ public class Policy extends PanacheEntityBase {
             readOnly = true,
             format = "yyyy-MM-dd hh:mm:ss.ddd",
             implementation = String.class)
+    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
```
josejulio, line 92: The `read_only` is on the sense of the json-string to/from java class. 
`read_only` here means this property will only be `read` (i.e. when converting from class to json) and `write_only` (not used in this PR) is when converting from json to class (i.e. write the class attribute).

I'm clarifying because is always confusing to me after a while. read/write is not applicable to the json string, but to the class itself.

src/main/java/com/redhat/cloud/policies/app/rest/PolicyCrudService.java
```diff
@@ -625,14 +629,21 @@ public Response validateCondition(@Valid @NotNull Policy policy) {
             @APIResponse(responseCode = "500", description = "Internal error")
     })
     @Parameter(name = "id", description = "UUID of the policy")
-    public Response validateName(@NotNull JsonString policyName, @QueryParam("id") UUID id) {
+    public Response validateName(@NotNull String jsonPolicyName, @QueryParam("id") UUID id) {
```
josejulio, line 632: I changed a bit here this part. 
We are reciving a json string. If we want to read  `my cool policy` we expect to receive the raw string `"my cool policy"` which is a json string (application/json) as opposed to plain text.

For some reason after the change from json-b to jackson this wasn't working correctly. So I'm using the ObjectMapper below to parse the string (which should be a string).

src/main/java/com/redhat/cloud/policies/app/auth/HeaderHelper.java
```diff
@@ -42,8 +42,8 @@ public static Optional<XRhIdentity> getRhIdFromString(String xRhIdHeader) {
         XRhIdentity rhIdentity;
         try {
             String json_string = new String(Base64.getDecoder().decode(xRhIdHeader));
-            rhIdentity = jsonb.fromJson(json_string, XRhIdentity.class);
-        } catch (JsonbException jbe) {
+            rhIdentity = objectMapper.readValue(json_string, XRhIdentity.class);
+        } catch (JsonProcessingException jpe) {
             return Optional.empty();
```
josejulio, line 47: It's worth to log this exception. This is happening when (for some reason) the identity header can't be parsed. It should succeed all the time unless there is a change that is affecting the parsing.

```java
import io.quarkus.logging.Log;

Log.error("Error while parsing the identity header", jpe);
```

I suppose it should be in a separate commit?

vkrizan: I've stumbled on this whilst testing. It should be logged. Let's do that separately. Thank you.

PR #564, opened 2023-08-02
feat: inventory host groups support (for history)

vkrizan:
* adds parsing of host groups from RBAC
* stores allowed host groups to user principal(s)
* adds filtering of PoliciesHistoryEntries based on allowed host groups

Out of scope: returning host groups on the API.

*Hint: review by commits*

RHINENG-1282


josejulio:
> _Hint: review by commits_

Are you suggesting to review each commit as a single unit of work?

src/main/java/com/redhat/cloud/policies/app/auth/RbacClient.java
```diff
@@ -27,9 +27,12 @@ public class RbacClient {
...
      * Catching and returning null would end up in the next call directly
      * return null from the cache without retrying the remote call.
+     *
+     * Inventory permissions are also queried to get the list of host groups
+     * that the user has access to.
      */
     @CacheResult(cacheName = "rbac-cache")
     RbacRaw getRbacInfo(String xrhidHeader) {
-        return rbac.getRbacInfo("policies", xrhidHeader);
+        return rbac.getRbacInfo("policies,inventory", 100, xrhidHeader);
```
josejulio, line 36: Just a side comment.
100 is a pretty big number to start with, but with users being able to create stuff (as i understand it's possible with the host groups) someone will probably top that number eventually.

src/main/java/com/redhat/cloud/policies/app/auth/models/Access.java
```diff
@@ -18,9 +18,48 @@
...
+            return new String[3];
+        }
+        return permission.split(":");
+    }
+
+    public boolean isInventoryHostsRead() {
+        String[] permissionFields = getPermissionFields();
+        return permissionFields[0].equals(INVENTORY_HOSTS_READ[0])
+            && (permissionFields[1].equals(INVENTORY_HOSTS_READ[1]) || permissionFields[1].equals(ANY))
+            && (permissionFields[2].equals(INVENTORY_HOSTS_READ[2]) || permissionFields[2].equals(ANY));
```
josejulio, line 45: This will throw a NullPointerException if `permission` is null as each element of the `permissionFields` will be null.

Use [Objects.equals](https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-) instead

You will need to include `java.util.Objects`

```suggestion
        return Objects.equals(permissionFields[0], INVENTORY_HOSTS_READ[0])
            && (Objects.equals(permissionFields[1], INVENTORY_HOSTS_READ[1]) || Objects.equals(permissionFields[1], ANY))
            && (Objects.equals(permissionFields[2], INVENTORY_HOSTS_READ[2]) || Objects.equals(permissionFields[2], ANY));
```

src/main/java/com/redhat/cloud/policies/app/auth/models/Access.java
```diff
@@ -18,9 +18,48 @@
 
 import java.util.List;
 
-public class Access {
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+public class Access extends RbacRawCommon {
```
josejulio, line 23: As I understand, `RbacRawCommon` only expose some constants. You don't need to inherit from it.

src/main/java/com/redhat/cloud/policies/app/auth/models/RbacRawCommon.java
```diff
@@ -0,0 +1,19 @@
+package com.redhat.cloud.policies.app.auth.models;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+public class RbacRawCommon {
+
+    @JsonIgnore
```
josejulio, line 7: If you decide not to inherit from this class, you can skip the `@JsonIgnore`s

src/main/java/com/redhat/cloud/policies/app/auth/models/Access.java
vkrizan: I had a common methods there at first. But, yeah it can be a static class with constants.

petrblaho (approved):
LGTM!

src/main/java/com/redhat/cloud/policies/app/auth/models/RbacRaw.java
```diff
@@ -38,6 +39,36 @@ public boolean canDo(String application, String resource, String operation) {
 
     }
 
+    /*
+     * Returns Host Group Ids if restricted.
+     * Only inventory host read permissions (incl. widldcards) are observed.
```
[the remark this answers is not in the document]
josejulio: That's widl!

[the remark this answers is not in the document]
vkrizan: It no longer is :-/

src/main/java/com/redhat/cloud/policies/app/auth/models/Access.java
vkrizan: I'll keep it as is.

src/main/java/com/redhat/cloud/policies/app/auth/models/Access.java
vkrizan: Updated. Thanks.

PR #575, opened 2023-09-22
Revert "Removing wildcard from SC Branch Build Check "

casey-williams-rh:
Pushed to wrong branch

Reverts RedHatInsights/policies-ui-backend#574


xxlhacker (approved):
👍

PR #585, opened 2023-10-12
fix: Update production server

josejulio:
Updates OpenAPI's production server to point to console.redhat.com instead of the old cloud.redhat.com


vkrizan (approved):
Thank you!

PR #606, opened 2024-01-05
fix: restartPolicy in OS template

vkrizan:
Fixes
spec.jobTemplate.spec.template.spec.containers[0].restartPolicy: Forbidden: may not be set for non-init container

RHINENG-6805


petrblaho (approved):
LGTM!

PR #631, opened 2024-02-09
fix: sortable policy fields

vkrizan:
The main purpose of this PR is to fix sorting (and filtering) of policies by column names as described in the API schemas. Fixes sorting by `last_triggered`.

To accomplish the above, this PR introduces `@QueryableColumn` annotation that replaces `SortableColumn` and `FilterableColumn` enums. The annotation holds info about the name of the queryable column and whether it can be used in filtering and sorting (defaulting to true). `ColumnGetter` then provides the ability to retrieve a column by name from a model class. The name of a queryable column is determined either by annotation, jakarta's `@Column.name`, or Java field name.  `ColumnGetter#get` returns data in form of a `ColumnInfo`, which adds a `fieldName` (Java field name).

Instance of a `ColumnGetter` with a specific model class, is used within `PagingUtils` to get and check available fields for sorting and filtering. Java field name is then used for HQL queries.

RHINENG-7824


petrblaho (approved):
LGTM!

PR #649, opened 2024-04-01
refactor(dev): update postgresql-rds image tag

Victoremepunto:
postgresql-rds "-1" images are being deprecated; moving to postgresql-rds:12 instead


vkrizan (approved):
Thank you

PR #656, opened 2024-04-10
Add HMS floor plan export

shaded-enmity:
This PR adds data export to the HMS bucket for analytics.

RHINENG-9419

/ping @vkrizan @skateman

### DIFF
--- a/.rhcicd/clowdapp.yaml
+++ b/.rhcicd/clowdapp.yaml
@@ -140,6 +140,39 @@ objects:
                         secretKeyRef:
                           name: policies-ui-backend-db
                           key: ${DB_SECRET_PASSWORD_KEY}
+  - apiVersion: metrics.console.redhat.com/v1alpha1
+    kind: FloorPlan
+    metadata:
+      name: policies-backend-hms
+    spec:
+      database:
+        secretName: ${FLOORIST_DB_SECRET_NAME}
+      objectStore:
+        secretName: ${FLOORIST_HMS_BUCKET_SECRET_NAME}
+      logLevel: ${FLOORIST_LOGLEVEL}
+      suspend: ${{FLOORIST_SUSPEND}}
+      disabled: ${{FLOORIST_DISABLED}}
+      queries:
+        - prefix: hms_analytics/policies/status
+          chunksize: 100000
+          query: >-
+            SELECT
+              P.id AS policy_id,
+              P.org_id,
+              P.is_enabled,
+              P.mtime AS policy_mtime,
+              P.ctime AS policy_ctime,
+              P.actions,
+              PH.host_id,
+              PH.ctime AS policy_hit_time
+            FROM
+              policy P
+            LEFT JOIN
+              policies_history AS PH
+            ON
+              P.id = uuid(PH.policy_id);
+
+
 parameters:
   - name: CLOUDWATCH_ENABLED
     description: Enable Cloudwatch (or not)
@@ -171,6 +204,22 @@ parameters:
   - name: ENV_NAME
     description: ClowdEnvironment name (ephemeral, stage, prod)
     required: true
+  - name: FLOORIST_SUSPEND
+    description: Disable Floorist cronjob execution
+    required: true
+    value: 'true'
+  - name: FLOORIST_DISABLED
+    description: Determines whether to build the Floorist Job.
+    value: 'false'
+  - name: FLOORIST_DB_SECRET_NAME
+    description: database secret name
+    value: policies-ui-backend-db
+  - name: FLOORIST_HMS_BUCKET_SECRET_NAME
+    description: HMS bucket secret name
+    value: floorist-bucket
+  - name: FLOORIST_LOGLEVEL
+    description: Floorist loglevel config
+    value: 'INFO'
   - name: IMAGE
     description: Image URL
     value: quay.io/cloudservices/policies-ui-backend


.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
                         secretKeyRef:
                           name: policies-ui-backend-db
                           key: ${DB_SECRET_PASSWORD_KEY}
+  - apiVersion: metrics.console.redhat.com/v1alpha1
+    kind: FloorPlan
+    metadata:
+      name: policies-backend-hms
+    spec:
+      database:
+        secretName: ${FLOORIST_DB_SECRET_NAME}
```
vkrizan, line 149: This is a first FloorPlan for this project, so these template parameters would need to be added.

shaded-enmity: Updated

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
...
+      disabled: ${{FLOORIST_DISABLED}}
+      queries:
+        - prefix: hms_analytics/policies/status
+          query: >-
+            SELECT
+              id,
+              name,
+              org_id,
+              is_enabled,
+              conditions,
```
vkrizan, line 163: `conditions` contain the expression that evaluates policies. This should not be part of the metrics, as it can contain private information.

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
...
+              mtime,
+              ctime,
+              actions,
```
vkrizan, line 166: At the moment `actions` only allow one value: `notification`. This might not be necessary to be collected.

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
...
+              JSON_AGG(policy_history.host_id) as hosts
+            FROM
+              policy
+            JOIN
+              policy_history
```
vkrizan, line 171: `s/policy_history/policies_history/`

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
...
+            ON
+              policy_history.policy_id = policy.id
```
vkrizan, line 173: Unfortunately `policy_id` is type of VARCHAR.
Use:
```
policy.id = uuid(policies_history.policy_id)
```

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
...
+            SELECT
+              id,
+              name,
+              org_id,
+              is_enabled,
+              conditions,
+              mtime,
+              ctime,
+              actions,
+              JSON_AGG(policy_history.host_id) as hosts
```
vkrizan, line 167: `s/policy_history/policies_history/`

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
...
+      objectStore:
+        secretName: ${FLOORIST_HMS_BUCKET_SECRET_NAME}
+      logLevel: ${FLOORIST_LOGLEVEL}
+      suspend: ${{FLOORIST_SUSPEND}}
+      disabled: ${{FLOORIST_DISABLED}}
+      queries:
+        - prefix: hms_analytics/policies/status
+          query: >-
+            SELECT
+              id,
```
vkrizan, line 159: Some of the columns are ambiguous. Please use their full canonical name (together with an alias). Don't forget to change it on `GROUP BY` as well.

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,41 @@ objects:
...
+              actions,
+              JSON_AGG(policy_history.host_id) as hosts
+            FROM
+              policy
+            JOIN
+              policy_history
+            ON
+              policy_history.policy_id = policy.id
+            GROUP BY
+              id;
```
vkrizan, line 175: This is not enough for the JSON_AGG. 
```
Error: ERROR: column "policy.name" must appear in the GROUP BY clause or be used in an aggregate function (SQLSTATE 42803)
```
It would likely need a list of all fields on the group by.

.rhcicd/clowdapp.yaml
vkrizan: Never mind, there is also a `NULL` value.

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,40 @@ objects:
...
+        secretName: ${FLOORIST_HMS_BUCKET_SECRET_NAME}
+      logLevel: ${FLOORIST_LOGLEVEL}
+      suspend: ${{FLOORIST_SUSPEND}}
+      disabled: ${{FLOORIST_DISABLED}}
+      queries:
+        - prefix: hms_analytics/policies/status
+          query: >-
+            SELECT
+              P.id,
+              P.name,
```
vkrizan, line 160: Name isn't necessary to be collected, or is it?

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,40 @@ objects:
...
+              P.org_id,
+              P.is_enabled,
+              P.mtime,
+              P.ctime,
+              P.actions,
+              JSON_AGG(PH.host_id) as hosts
```
vkrizan, line 166: There are records over 1000. I'm wondering whether we should just flatten this:
```
            SELECT
              P.id AS policy_id,
              P.org_id,
              P.is_enabled,
              P.mtime AS policy_mtime,
              P.ctime AS policy_ctime,
              P.actions,
              PH.host_id,
              PH.ctime AS policy_hit_time
            FROM
              policy P
            LEFT JOIN
              policies_history AS PH
            ON
              P.id = uuid(PH.policy_id);
```

This one adds policy hit time and removes policy name.

.rhcicd/clowdapp.yaml
```diff
@@ -140,6 +140,40 @@ objects:
...
+      database:
+        secretName: ${FLOORIST_DB_SECRET_NAME}
+      objectStore:
+        secretName: ${FLOORIST_HMS_BUCKET_SECRET_NAME}
+      logLevel: ${FLOORIST_LOGLEVEL}
+      suspend: ${{FLOORIST_SUSPEND}}
+      disabled: ${{FLOORIST_DISABLED}}
+      queries:
+        - prefix: hms_analytics/policies/status
+          query: >-
```
vkrizan, line 157: I think we should add some `chunksize: ` as the default is 1000. Something that would be within the memory limits, otherwise we would need to increase them. I'd start with `chunksize: 100000`.

shaded-enmity: Ack, AFAIU this will reduce the number of exported files so that they compress much better?

.rhcicd/clowdapp.yaml
shaded-enmity: yeah I don't think so

shaded-enmity:
@vkrizan Thanks for the suggestions, updated, PTAL.

vkrizan (approved):
Thank you Pavel!

PR #659, opened 2024-04-26
fix(build_deploy): only create sc tag when building sc branch

casey-williams-rh:
It has been observed  that the `build_deploy` script is set up so that when the SC build Job runs, it write a new image over the original `image_tag` on which the commit is based. 

This update fixes that issue.


vkrizan (approved):
Thank you

PR #680, opened 2024-07-25
fix(tests): use specific postgres container image

vkrizan:
Fixes test runs with testcontainers with podman by specifying the image with repository.


petrblaho (approved):
LGTM

PR #690, opened 2024-09-11
Adding floorist plugin tests to pr_check.sh

prichard77:


prichard77:
RHINENG-12132

.rhcicd/pr_check.sh
```diff
@@ -9,8 +9,10 @@ export IMAGE="quay.io/cloudservices/policies-ui-backend"
 export DEPLOY_TIMEOUT="600"
 
 # IQE plugin config
-export IQE_PLUGINS="policies"
-export IQE_MARKER_EXPRESSION="policies_api_smoke"
+#export IQE_PLUGINS="policies,floorist"
+export IQE_PLUGINS="floorist"
+#export IQE_MARKER_EXPRESSION="policies_api_smoke or floorist_smoke"
+export IQE_MARKER_EXPRESSION="floorist_smoke"
```
vkrizan, line 15: This turns off (removes) policies testing.

vkrizan (changes requested):

.rhcicd/pr_check.sh
prichard77: yes. I was just trying to troubleshoot. I was having issues with the image not having the plugin. I was just seeing what happened when I just tried to run the floorplan tests. This will not be in the final version. ALSO, Mike Shriver has shown me a new approach. It involves more updates to additional file that this, but will give us more flexibility.

.rhcicd/pr_check_fp.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+set -exv
+
+# Clowder config
+export APP_NAME="policies"
+export COMPONENT_NAME="policies-ui-backend"
+export IMAGE="quay.io/cloudservices/iqe-tests:latest"
```
mshriver, line 8: If floorist is getting installed in `iqe-tests:policies` you can still use it here.

.rhcicd/pr_check_fp.sh
```diff
@@ -0,0 +1,38 @@
...
+export DEPLOY_TIMEOUT="600"
+
+# IQE plugin config
+export IQE_PLUGINS="floorist"
```
vkrizan, line 12: Can we have this together with the main policies plugin and have it all in one PR check (file)?

.rhcicd/pr_check_fp.sh
prichard77: The iqe-floorist-plugin does not seem to be part of the :policices tag. I could add the floorist plugin as a dependency in the policices plugin and then use the :policices tag.

prichard77: I updated the MR to use iqe-tests:policies after adding it to the image.

.rhcicd/pr_check_fp.sh
vkrizan: What gating properties are you concerned about?

Can you elaborate on what difficulties there were when running 2 plugins with one PR check?

.rhcicd/pr_check_fp.sh
```diff
@@ -0,0 +1,38 @@
...
+export IQE_FILTER_EXPRESSION=""
+export IQE_CJI_TIMEOUT="30m"
+
+# Bonfire init
+CICD_URL=https://raw.githubusercontent.com/RedHatInsights/bonfire/master/cicd
+curl -s $CICD_URL/bootstrap.sh > .cicd_bootstrap.sh && source .cicd_bootstrap.sh
+
+# Build the image and push to Quay
+export DOCKERFILE=src/main/docker/Dockerfile.jvm
+source $CICD_ROOT/build.sh
```
vkrizan, line 23: If this needs to be standalone, why does it (re)build the image? This needs to be dependent on the main PR check.

.rhcicd/pr_check_fp.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Deploy on ephemeral
+export COMPONENTS_W_RESOURCES="policies-engine"
+source $CICD_ROOT/deploy_ephemeral_env.sh
```
vkrizan, line 27: Running this standalone would require another environment to be spawn, which increases resource utilization.

.rhcicd/pr_check_fp.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Run smoke tests with ClowdJobInvocation
+source $CICD_ROOT/cji_smoke_test.sh
+
+# Until test results produce a junit XML file, create a dummy result file so Jenkins will pass
+mkdir -p $WORKSPACE/artifacts
+cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
+<testsuite tests="1">
+    <testcase classname="dummy" name="dummytest"/>
```
vkrizan, line 36: For next improvment, it would be good to provide a proper test report.

.rhcicd/pr_check_fp.sh
vkrizan: I recommend to extend the original PR check. If the IQE env variables cannot be combined, then we can run first (original) tests, then change the IQE env vars (that would set the floorist IQE plugin to run) and then run the `cji_smoke_test.sh` (or similar) second time.

.rhcicd/pr_check_fp.sh
prichard77: I'll be happy to look into this more. However, at first glance it appears that this is related to the unit tests. We get results for all of the iqe-tests. These show up as (root) in the output.

prichard77:
The changes here were implemented in https://github.com/RedHatInsights/policies-ui-backend/pull/693 . I am closing.

PR #692, opened 2024-09-17
Add floorist to pr_check using a single cli

prichard77:
trying this one more time using one cji before attempting a solution that requires 2.


prichard77:
I just wanted to try this once again before I tried using 2 cjis.  I am not sure where this pr_check gets it's iqe-tests image containing the policies plugin. It must be using a plugin image somewhere, but I added the floorist plugin to the iqe-policies-plugin and it is not getting found in these runs.

prichard77:
The changes here were implemented in https://github.com/RedHatInsights/policies-ui-backend/pull/693 . I am closing.

PR #693, opened 2024-09-17
Add floorist plugin tests using a second cji

prichard77:
RHINENG-12132


prichard77:
https://issues.redhat.com/browse/RHINENG-12132

.rhcicd/pr_check.sh
```diff
@@ -36,3 +36,13 @@ cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
     <testcase classname="dummy" name="dummytest"/>
 </testsuite>
 EOF
+
+# adding floorist tests.
```
mshriver, line 40: Consider replacing the above block that is dummying the artifact file, since you don't need it when your tests are actually added and running. 

It's a placeholder for when the CJI isn't running anything yet.

.rhcicd/pr_check.sh
```diff
@@ -36,3 +36,13 @@ cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
...
+export COMPONENT_NAME="policies-engine"
```
vkrizan, line 41: Lets not override component name. Is this to hack this around to have a second CJI run?

.rhcicd/pr_check.sh
```diff
@@ -36,3 +36,13 @@ cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
...
+# I also need to pass in FLOORPLAN_2_TEST.
+export IQE_ENV_VARS="FLOORPLAN_TO_TEST=policies-backend-hms"
+export IQE_PLUGINS="floorist"
+export IQE_MARKER_EXPRESSION="floorist_smoke"
+export IMAGE="quay.io/cloudservices/iqe-tests:floorist"
```
vkrizan, line 46: Don't override `IMAGE`. That env var sets the service that runs, not a test image. The service is policies-ui-backend.

.rhcicd/pr_check.sh
```diff
@@ -36,3 +36,13 @@ cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
...
+# Run smoke tests with ClowdJobInvocation
+source $CICD_ROOT/cji_smoke_test.sh
```
vkrizan, line 48: No new line at the end of the file.

.rhcicd/pr_check.sh
prichard77: This is how I was able to get the pod to deploy. This is what I had to use to get the pod to deploy when I was running  deploy-iqe-cji manually. I do not believe it will deploy otherwise.

.rhcicd/pr_check.sh
prichard77: This controls what image is used for the pod running iqe as I understand. The pod will not deploy using policies-ui-backend.

.rhcicd/pr_check.sh
prichard77: I'll add one.

.rhcicd/pr_check.sh
mshriver: This has to be set to the component that IQE will use in the CJI to lookup the clowdenv configmap

.rhcicd/pr_check.sh
mshriver: you want to use `IQE_IMAGE_TAG`

mshriver: https://github.com/RedHatInsights/cicd-tools/blob/main/cji_smoke_test.sh#L8

.rhcicd/pr_check.sh
prichard77: You will also notice that the existing line 26 has this: export COMPONENTS_W_RESOURCES="policies-engine".

.rhcicd/pr_check.sh
prichard77: I replaced IMAGE with IQE_IMAGE_TAG.

mshriver: ack - it should just be the tag though, not the whole image URI. 

```  -i, --image-tag TEXT          image tag to use for IQE pod```

Comment (prichard77) [reply .rhcicd/pr_check.sh]: Updating...

File: .rhcicd/pr_check.sh (hunk 2 above)
Comment (prichard77) [reply .rhcicd/pr_check.sh:36]: I am testing removing line 41 in a other PR. I'll make the change here if it passes.

Comment (prichard77) [reply .rhcicd/pr_check.sh:36]: I tried removing line 41 but the tests time out attempting to deploy via cji. I ran into this when I was deploying and testing in ephemeral manually. You will notice that in a successful run here we use "bonfire deploy-iqe-cji policies-engine" to deploy resources to run the floorist tests. When I remove line 41 (export COMPONENT_NAME="policies-engine")    "bonfire deploy-iqe-cji policies-ui-backend" is used and the deployment times out. You can see this here:  https://ci.ext.devshift.net/job/RedHatInsights-policies-ui-backend-pr-check/705/console

Comment (vkrizan) [reply .rhcicd/pr_check.sh:36]: This needs to be set to the same component name, otherwise it won't work as it wouldn't get the right credentials to the database. If we're going to have this running in other projects, not all have two (or more) components. So we don't have that luxury to hack it around with using a different component.

Can you please check what is causing the timeout? I see that it does want to reuse the same test pod. Is there a way to force it to use a new one with the new name?

Comment (vkrizan) [reply .rhcicd/pr_check.sh:36]: @prichard77 try to utilize option `-c` (`--cji-name`) for the CJI deploy command. Example: `--cji-name floorist-iqe`

File: .rhcicd/pr_check.sh [hunk 5]
```diff
@@ -31,8 +31,13 @@ source $CICD_ROOT/cji_smoke_test.sh
...
-cat << EOF > ${WORKSPACE}/artifacts/junit-dummy.xml
-<testsuite tests="1">
-    <testcase classname="dummy" name="dummytest"/>
-</testsuite>
-EOF
+
+# adding floorist tests.
+export COMPONENT_NAME="policies-engine"
+# I also need to pass in FLOORPLAN_2_TEST.
+export IQE_ENV_VARS="FLOORPLAN_TO_TEST=policies-backend-hms"
```
vkrizan, line 38: Would we support more than one floorplan to test? For example by providing more plan names separated by comma?

prichard77: we could support multiple floorplans. However, I'd like to do this in another story/task for the floorist plugin. It should not be a huge effort. I actually thought about implementing this initially but we did not state we needed it.

prichard77:
I created an MR to update cji_smoke_tests.sh and was able to do a small bit of testing. When this update gets merged, we should be able to change the cji name and hopefully get closer to merging. I will go ahead and add the needed update to support this change as it should be benign. I only need to add an env var.

.rhcicd/pr_check.sh
vkrizan: I think we've discussed this as one of the requirements, as services already use two floorplans. Anyways, we can extend it later.

.rhcicd/pr_check.sh
prichard77: I added support to pass in cji name and also I added support for multiple floorplans.

.rhcicd/pr_check.sh
prichard77: I added support for multiple floorplans: https://gitlab.cee.redhat.com/insights-qe/iqe-floorist-plugin/-/merge_requests/24

.rhcicd/pr_check.sh
vkrizan: The `COMPONENT_NAME` is still wrong. Please remove this line.

mshriver:
```
(.iqe-env) ➜  iqe bonfire deploy-iqe-cji policies-ui-backend --cji-name iqe-floorist-tests --marker floorist_smoke --plugins floorist --env-var "FLOORPLANS_TO_TEST=policies-backend-hms" --image-tag floorist --env clowder_smoke --namespace $NAMESPACE

================= 1 passed, 1 deselected, 4 warnings in 0.63s ==================

(.iqe-env) ➜  iqe bonfire deploy-iqe-cji policies-engine --cji-name iqe-floorist-tests-engine --marker floorist_smoke --plugins floorist --env-var "FLOORPLANS_TO_TEST=policies-backend-hms" --image-tag floorist --env clowder_smoke --namespace $NAMESPACE

================= 1 passed, 1 deselected, 4 warnings in 0.66s ==================
```

Neither is wrong. This controls the clowdapp that owns the CJI, not a specific component cdappconfig that IQE will have access to, or what the IQE pod is testing against.

.rhcicd/pr_check.sh
```diff
@@ -31,8 +31,17 @@ source $CICD_ROOT/cji_smoke_test.sh
...
-<testsuite tests="1">
-    <testcase classname="dummy" name="dummytest"/>
-</testsuite>
-EOF
+
+# adding floorist tests.
+<<<<<<< Updated upstream
+export COMPONENT_NAME="policies-engine"
+=======
+>>>>>>> Stashed changes
```
vkrizan, line 39: This looks like an unresolved conflict.

prichard77: Thanks for pointing this out. I am not sure what happened to get this code inserted. Something I have not seen before with git stash.

prichard77: I am guessing I stashed something that I then changed upstream before I popped it and added to commit.

vkrizan (approved):
Thank you!

PR #694, opened 2024-09-18
Try no new comp name

prichard77:


prichard77:
I am just testing something here. I had this set as a draft, but did not want that to stop any tests from being run.

vkrizan:
Superseded by #693

PR #716, opened 2024-10-24
Revert to UBI8 images

patchkez:
RHINENG-13893
Please merge only when Konflux supports UBI8 (Monday 28th of Oct).


skateman:
If konflux is green, this can go in!

vkrizan (approved):

PR #721, opened 2024-11-05
fix(license): add NOTICE file specifying Red Had as the owner

RoamingNoMaD:
As per my email thread, I'd like to comply with the license appendix and add a NOTICE file specifying the owner of the copyright.


petrblaho (approved):
LGTM!

PR #735, opened 2024-12-18
fix: do not use public pg schema

vkrizan:
In order to have Policies running on different PG schemas, the migrations should not use fully qualified table names with the "public" schema.  Rather names of tables should stay schema-less, to leverage default schema set by search_path.


petrblaho (approved):
LGTM

PR #742, opened 2025-01-08
fix(db): repair already applied migrations

vkrizan:
Runs repair on already applied migrations on start to match migrations that have been changed retroactively.

Ref: https://quarkus.io/version/3.15/guides/flyway


vkrizan:
Supersedes #741

fellipeh (approved):
lgtm

PR #743, opened 2025-01-08
Change CPU/MEMORY limits for clowdapp - ADR46

patchkez:
This change is implementing changes recommended by 'Trusted Resource Configuration (ADR-46) Implementation' to limit CPU/MEM requests/limits in ephemeral environment.


vkrizan (approved):
This looks good to me. I'll leave the merge to @fellipeh. Thank you.

PR #839, opened 2025-09-08
fix(ci): Enable renovate updates for tekton pipelines anytime

patchkez:
## Summary by Sourcery

CI:
- Enable Renovate updates for Tekton pipeline dependencies outside of the previously configured schedule


vkrizan (approved):
Considering that these updates would most of the time update the reference to common pipelines, having those anytime can shorten the time.

